PR #55, opened 2021-05-03
Add wrapper for Confluence gallery macro

Vertganti:
The supported parameters are 'title', 'include' and 'exclude'. I tried implementing the 'page' parameter, but it does not work yet. I have stated this on the macro page. 

The Confluence macro has to be prefixed during import in order to use this wrapper, as there already is a gallery macro in XWiki (which this wrapper uses behind the scenes).


Vertganti:
Fixed the page parameter by using the document title for the query.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+    <property>
+      <async_context/>
+    </property>
+    <property>
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <code>{{velocity}}
+## Get the parameters and put them in lists if necessary
+#set($title = $wikimacro.parameters.title)
```
acotiuga, line 326: In order to follow the best practices and keep code clean and consistent, we should always use a blank space in the velocity statements using `#set ()`, `#if ()`, `#foreach ()`. I see that you used this in few places, so it would be nice to have it everywhere.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+#set($includeString = $wikimacro.parameters.include)
+#set($excludeString = $wikimacro.parameters.exclude)
+#set($pagesString = $wikimacro.parameters.page)
+#set($include = $stringtool.split($includeString, ","))
```
acotiuga, line 330: Here, you could improve the code using `#set ($include = $includeString.split(','))` as a shorter form
Also, if there is no velocity variable in the string, we should use single quote `'` to wrap the string

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+#set($exclude = $stringtool.split($excludeString, ","))
+#set($pages = $stringtool.split($pagesString, ","))
+## Display the gallery title if it is set (newlines are needed for the following macro calls to be standalone)
+#if(!$stringtool.isEmpty($title))
+    ==$title==
+    
+
```
acotiuga, line 337: I did some tests and I'm not sure is necessary to have these empty lines, at least not in the final code. Maybe at some point, for debugging purposes, you printed some data that was breaking the macro. It works for me now. Please test again and if you find out that these are not needed, remove them.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+#set($title = $wikimacro.parameters.title)
+#set($includeString = $wikimacro.parameters.include)
+#set($excludeString = $wikimacro.parameters.exclude)
+#set($pagesString = $wikimacro.parameters.page)
+#set($include = $stringtool.split($includeString, ","))
+#set($exclude = $stringtool.split($excludeString, ","))
+#set($pages = $stringtool.split($pagesString, ","))
+## Display the gallery title if it is set (newlines are needed for the following macro calls to be standalone)
+#if(!$stringtool.isEmpty($title))
+    ==$title==
```
acotiuga, line 335: All the code should be indented with 2 spaces and not 4. Otherwise put with a tab of 2 spaces and not a tab of 4 spaces. Please fix the code indentation.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+## add the current page to the page list if no pages were given
+#if($stringtool.isEmpty($pagesString))
+    #set ($pages = [$doc.getTitle()])
+#end
+## Collect the attachments to display
+#set ($attachments = [])
+#foreach($page in $pages)
+    ## Resolve page and put attachments in temporary list
+    #set ($tempAttachments = [])
+    #set($query = $services.query.xwql("where doc.title = :name").bindValue('name', ''+${page}))
```
acotiuga, line 348:
```suggestion
    #set ($query = $services.query.xwql('where doc.title = :name').bindValue('name', $page))
```

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+    #set($pageStrings = $query.execute())
+    #if ($pageStrings.size() &gt; 0)
+        #set($pageDocument = $xwiki.getDocument($pageStrings[0]))
```
acotiuga, line 351: This code doesn't seem to handle multiple pages in the `page` parameter while the description says `Pages containing the images to display.` Which of these is correct?

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+        #set($discard = $tempAttachments.addAll($pageDocument.getAttachmentList()))
+        ## iterate over remaining attachments and add full reference to attachment list
+        #foreach($attachment in $tempAttachments)
+            #if($attachment.isImage())
+                ## If includes or excludes are set, remove elements from temporary attachment list
```
acotiuga, line 356: The comment is not clear here as there is no removal operation after it.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+## Get the parameters and put them in lists if necessary
+#set($title = $wikimacro.parameters.title)
+#set($includeString = $wikimacro.parameters.include)
+#set($excludeString = $wikimacro.parameters.exclude)
+#set($pagesString = $wikimacro.parameters.page)
+#set($include = $stringtool.split($includeString, ","))
+#set($exclude = $stringtool.split($excludeString, ","))
+#set($pages = $stringtool.split($pagesString, ","))
+## Display the gallery title if it is set (newlines are needed for the following macro calls to be standalone)
+#if(!$stringtool.isEmpty($title))
```
acotiuga, line 334: In Velocity it's simpler to check strings for null and emptiness by using `#if ("$!title" != '')`, meaning the `$title` variable is not null and is not empty.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+    #set($pageStrings = $query.execute())
+    #if ($pageStrings.size() &gt; 0)
+        #set($pageDocument = $xwiki.getDocument($pageStrings[0]))
+        #set($discard = $tempAttachments.addAll($pageDocument.getAttachmentList()))
+        ## iterate over remaining attachments and add full reference to attachment list
+        #foreach($attachment in $tempAttachments)
+            #if($attachment.isImage())
+                ## If includes or excludes are set, remove elements from temporary attachment list
+                #if(!$stringtool.isEmpty($includeString))
+                    #if ($include.contains($attachment.filename))
```
acotiuga, line 358: Why not grouping these 2 if statements in `#if(!$stringtool.isEmpty($includeString) && $include.contains($attachment.filename)))`
Even better, you could check if the `$includeString` is empty when assigning `$include`, then check only if the $include contains the filename:
```
#set ($includeString = $wikimacro.parameters.include)
...
#if ("$!includeString" != '')
  #set ($include = $includeString.split(','))
#end
...
#if ($include.contains($attachment.filename))
  #set ($discard = $attachments.add($pageDocument + '@' + $attachment.filename))
#end
```

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
```diff
@@ -0,0 +1,723 @@
...
+                        #set($discard = $attachments.add($pageDocument+'@'+$attachment.filename))
+                    #end
+                #elseif(!$stringtool.isEmpty($excludeString))
+                    #if (!$exclude.contains($attachment.filename))
+                        #set($discard = $attachments.add($pageDocument+'@'+$attachment.filename))
```
acotiuga, line 363: Same as above.

acotiuga (changes requested):
The functionality works except is not clear if it should work with multiple pages or single page in the page parameter. The rest observations are purely code style which I would like to be improved.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
Vertganti: I had some problems with a standalone macro, but I cannot reproduce them right now. As you say it might have been other debugging prints breaking the macro. I will remove the lines.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
Vertganti: I had to use four spaces because I copy the code of all developed macros into our internal documentation on GitLab. If the code only has two spaces the GitLab editor breaks indentation when using Tab/Shift+Tab to change the indentation. In this case I will adapt the indentation again for the PR.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
Vertganti: It does handle multiple pages. The page parameter ($pagesString) is split on commas ($pages) and then the macro iterates over the entries in the $pages list.

acotiuga: Indeed, the code is getting all the documents having a given title, hoping there is only one.

src/main/resources/Confluence/Macros/ConfluenceGallery.xml
Vertganti: Will velocity simply skip the `#if ($include.contains($attachment.filename))` clause when `$include` is not defined? Based on Java I would expect NullPointer issues, but my Velocity knowledge is very limited.

acotiuga: Yes, it will silently fail, meaning that if `$include` is not defined, the evaluation result in `false`.

Vertganti: Okay, thanks. On rereading the code I remember that I not group the ifs on purpose, as I only wanted to handle excludes if no includes are specified (includes taking precedence over excludes). If includes are specified everything else is excluded anyway, so no need to extra handle excludes (unless someone wants to exclude an included image, which I see no sense in). Excludes are then only handled if there are no includes. If I group the statements there may be the case that the include string is not empty, but the current attachment is not in it. In that case the macro would then proceed to (unnecessarily) check if the image should be excluded. The same would happen for the second approach you posted. It does not make that much of a difference, I just wanted to avoid unnecessary calls. If you want I can rewrite it anyway.

acotiuga (approved):

acotiuga:
Thank you for your contribution!

PR #61, opened 2021-08-16
Implement details and detailssummary macros (#20)

Vertganti:
This PR adds implementations for the details (page properties) and detailssummary (page properties report) confluence macros. The detailssummary implementation is very basic and will only summarize the details macros of all subpages where used. Parts of the code were provided by @ldubost .


Vertganti:
I don't know why the commits for the gallery macro still show up in the PR. Only the last three commits are actually related.

src/main/resources/Confluence/Macros/DetailsSummary.xml
```diff
@@ -0,0 +1,420 @@
...
+
+// debug handling
+debug = ""
+
+def addDebug(msg) {
+  if (request.debug)
+    debug += msg + "\n";
+}
+
+// define functions for parsing details macros, based on code by Ludovic Dubost
```
raphj, line 268: can you give a link to this code?

src/main/resources/Confluence/Macros/DetailsSummary.xml
```diff
@@ -0,0 +1,420 @@
...
+      <async_cached>0</async_cached>
+    </property>
+    <property>
+      <async_context/>
+    </property>
+    <property>
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <code>{{groovy}}
```
raphj, line 254: It would be best not to rely on groovy because it requires programming rights for the user installing the extension, which is not necessarily the case.

Instead, this would be best implemented as a Java script service in pro-macros-api.

Another thing: this code and the code in the DetailsSummaryResult page is duplicated, they should be merged together and the script component will allow doing this.

src/main/resources/Confluence/Macros/DetailsSummary.xml
```diff
@@ -0,0 +1,420 @@
...
+  }
+}
+
+// store variable in the xcontext so we can use it in velocity
+xcontext.put("columns", columns)
+{{/groovy}}
+
+{{velocity}}
+## get the variable back from the context
+#set ($columns = $xcontext.get("columns"))
```
raphj, line 377: With the Groovy code turned into a script component, this would become a method call and we will be able to avoid storing the result in the context.

src/main/resources/Confluence/Macros/DetailsSummary.xml
```diff
@@ -0,0 +1,420 @@
...
+#foreach ($column in $columns)
+  #set ($discard = $columnsProperties.put($column, { "type" : "none", "html" : "true" }))
+  ##set ($discard = $columnsProperties.put($column, { "type" : "text" }))
+#end
+## set the page to query for results and pass the page from which the macro is called as an extra parameter called 'source'
+#set ($options = {
+  "resultPage":"Confluence.Macros.DetailsSummaryResult",
+  "extraParams":"source=$escapetool.url($doc)"
+})
+#livetable("testtable" $columns $columnsProperties $options)
```
raphj, line 389: I'd appreciate a more explicit livetable ID, like "detailssummary". What's more, we may need to ensure uniqueness, maybe by appending some counter. Counter that could be put in the context.

src/main/resources/Confluence/Macros/DetailsSummaryResult.xml
```diff
@@ -0,0 +1,257 @@
...
+  <contentAuthor>xwiki:XWiki.Admin</contentAuthor>
+  <date>1624971517000</date>
+  <contentUpdateDate>1624971517000</contentUpdateDate>
+  <version>1.1</version>
+  <title/>
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>true</hidden>
+  <content>{{groovy}}
```
raphj, line 42: Duplicated code that we will need to deduplicate. After this duplication, involving the creation of a new script service, the live table results will be able to use this script service instead.

src/main/resources/Confluence/Macros/DetailsSummaryResult.xml
```diff
@@ -0,0 +1,257 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<xwikidoc version="1.4" reference="Confluence.Macros.DetailsSummaryResult" locale="">
+  <web>Confluence.Macros</web>
+  <name>DetailsSummaryResult</name>
```
raphj, line 25: So far, I've always seen custom live table result pages called  `SomethingLiveTableResults`. I don't know if this is a hard rule but I'd prefer this page to follow this convention.

`DetailsSummaryLiveTableResults` would be a good fit.

raphj (changes requested):
Thanks for your contribution and sorry it took us that much time to look into it!

This work is very valuable, we are looking to merge it soon. I'm also looking into implementing CQL filters on my side and your work is a very good starting point.

Some change will be needed before we can merge. Do you still want to work on this? Otherwise, I can take over for any or all these comments, depending on what you can/want to work on.

Let me know if you want to / are able to work on this in the following days, or any though you might have on this work or my feedback.

Thanks again for this contribution!

raphj:
Ah, one more thing: the directory layout of the repository has changed, this branch will need to be rebased on or merged with the new structure.

Vertganti:
These macro were a requirement of another department which in the end chose to stay with Confluence. In our current XWiki setup we do not use the macros so I do not intend to continue working on this code. Also, I haven't worked with this codebase for the last 1.5 years, so I would probably have a much harder time than you. Thanks for taking over and feel free to continue work on it. Your comments all look fine to me.

src/main/resources/Confluence/Macros/DetailsSummary.xml
Vertganti: It was some private code in an internal XWiki GitLab repository so I cannot provide the link, sorry. I mentioned it in the comment because I didn't want to take credit for code which is not mine.

snazare:
@raphj is it ok to close this PR as the two macros were implemented?

PR #63, opened 2022-03-17
Add Team Macro

### LINKED ISSUE #86: Implement the Team macro
This macro will display a team (from a XWiki Group) using their avatars, pretty much like what can be seen at the right side of the screen (the "People" section) on a github organization page like https://github.com/xwikisas.

raphj:
This pull request adds a new macro to XWiki Pro Macros: the team macro, which displays a team in a list of avatar, as can be seen in many tools, like GitHub organizations for instance.

Closes #86


src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+    $tagwhere
+  ORDER BY doc.name
+")
+#set($list = $services.query.hql($sql).execute())
+{{html wiki=false clean=false}}
+&lt;div class="xwikiteam #if($showUsernames)xwikiteam-usernames-displayed #else xwikiteam-usernames-hidden #end"&gt;
+  #if ($list.size() &gt; 0)
+    &lt;ul class="xwikiteam-ul"&gt;
+      #foreach($user in $list)
+        #avatar("xwiki:${user}" $size $disableLetterAvatars $letterAvatarBgColor $letterAvatarFontColor)
```
mflorea, line 801: You should **resolve** the local user reference relative to the current wiki (I guess, since the query above is executed in the current wiki, which means the team macro is looking only for users defined in the current wiki), instead of computing the absolute user reference by concatenating the wiki reference (not to mention that this won't work correctly inside a subwiki, that has a different reference than 'xwiki').

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+          #end
+        #end
+      #end
+      #if($pictureURL == "" &amp;&amp; !$disableLetterAvatars)
+        #set($picture = "#avatarInitials($displayUser $size $letterAvatarBgColor $letterAvatarFontColor)")
+      #else
+        #if($pictureURL == "")
+          #set($pictureURL = $xwiki.getAttachmentURL('XWiki.XWikiUserSheet', 'noavatar.png'))
+        #end
+        #set($picture = "&lt;img class='xwikiteam-avatar' src='$pictureURL' alt='$displayUser' title='$displayUser' width='$size' height='$size' /&gt;")
```
mflorea, line 696: Missing XML escaping.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+      #end
+      #set($spanName = "&lt;span class='xwikiteam-username'&gt;$displayUser&lt;/span&gt;")
```
mflorea, line 698: Missing XML escaping (we don't want XSS through user name).

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+  #if ($list.size() &gt; 0)
+    &lt;ul class="xwikiteam-ul"&gt;
+      #foreach($user in $list)
+        #avatar("xwiki:${user}" $size $disableLetterAvatars $letterAvatarBgColor $letterAvatarFontColor)
+      #end
+      #if(!$disableTools)
+        &lt;li class="xwikiteam-tools-link" hidden="hidden"&gt;
+          &lt;a href="#" role="button" title="${services.localization.render('rendering.macro.team.content.options')}"&gt;
+            &lt;span class="fa fa-wrench"&gt;&lt;/span&gt;
+            &lt;span class="sr-only"&gt;${services.localization.render('rendering.macro.team.content.options')}&lt;/span&gt;
```
mflorea, line 807: Missing XML escaping.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+        &lt;input
+          class="xwikiteam-show-username-checkbox"
+          type="checkbox"
+          #if($displayUsernames)checked="checked"#end
+        /&gt;
+        ${services.localization.render('rendering.macro.team.content.showUsernames')}
+      &lt;/label&gt;
+    &lt;/div&gt;
+  #else
+    &lt;p&gt;${services.localization.render('rendering.macro.team.content.noUsers')}&lt;/p&gt;
```
mflorea, line 824: Missing XML escaping, here and above.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+      <id>team</id>
+    </property>
+    <property>
+      <name>XWiki Team Image Display</name>
+    </property>
+    <property>
+      <priority/>
+    </property>
+    <property>
+      <supportsInlineMode>1</supportsInlineMode>
```
mflorea, line 855: I see it always generates DIVs so I don't see how it can be used inline, e.g. inside a paragraph.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+      <defaultCategory>content</defaultCategory>
+    </property>
+    <property>
+      <description>This macro displays the team members which have a specific tag</description>
+    </property>
+    <property>
+      <id>team</id>
+    </property>
+    <property>
+      <name>XWiki Team Image Display</name>
```
mflorea, line 849: Why not simply "Team". We're no prefixing our macros with "XWiki" and I'm not sure what "team image" means (e.g. "how do you specify the team image?").

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+    </property>
+    <property>
+      <priority/>
+    </property>
+    <property>
+      <supportsInlineMode>1</supportsInlineMode>
+    </property>
+    <property>
+      <visibility>Global</visibility>
```
mflorea, line 858: It should be visible only at wiki level, otherwise it has to be installed / saved by someone that has Programming Rights in order to work.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+        <name>type</name>
+        <number>5</number>
+        <prettyName>Parameter type</prettyName>
+        <size>60</size>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </type>
+    </class>
+    <property>
+      <defaultValue></defaultValue>
```
mflorea, line 1002: Missing default value. I don't think '' is the default size, is it?

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
```
mflorea, line 1080: Missing default value.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
```
mflorea, line 1158: Missing default value.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
```
mflorea, line 1236: Is '' the default color?

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
```
mflorea, line 1314: Is '' the default font color?

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
```
mflorea, line 1392: Missing default value.

src/main/resources/XWiki/Macros/TeamConfiguration.xml
```diff
@@ -0,0 +1,420 @@
...
+  <defaultLanguage>en</defaultLanguage>
+  <translation>0</translation>
+  <creator>XWiki.RaphaelJakse</creator>
+  <creationDate>1647417931000</creationDate>
+  <parent>WebHome</parent>
+  <author>XWiki.RaphaelJakse</author>
+  <contentAuthor>XWiki.RaphaelJakse</contentAuthor>
+  <date>1647534216000</date>
+  <contentUpdateDate>1647417931000</contentUpdateDate>
+  <version>12.5</version>
```
mflorea, line 15: Missing ``mvn xar:format``.

src/main/resources/XWiki/Macros/TeamConfiguration.xml
```diff
@@ -0,0 +1,420 @@
...
+  <title/>
+  <comment/>
+  <minorEdit>true</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>false</hidden>
```
mflorea, line 20: Should be hidden.

src/main/resources/XWiki/Macros/TeamConfigurationClass.xml
```diff
@@ -0,0 +1,134 @@
...
+  <defaultLanguage/>
+  <translation>0</translation>
+  <creator>XWiki.RaphaelJakse</creator>
+  <creationDate>1647417123000</creationDate>
+  <parent>WebHome</parent>
+  <author>XWiki.RaphaelJakse</author>
+  <contentAuthor>XWiki.RaphaelJakse</contentAuthor>
+  <date>1647533770000</date>
+  <contentUpdateDate>1647417124000</contentUpdateDate>
+  <version>1.39</version>
```
mflorea, line 15: Missing ``mvn xar:format``.

src/main/resources/XWiki/Macros/TeamConfigurationClass.xml
```diff
@@ -0,0 +1,134 @@
...
+      <number>4</number>
+      <numberType>integer</numberType>
+      <prettyName>Size of the avatars, in pixels</prettyName>
+      <size>30</size>
+      <unmodifiable>0</unmodifiable>
+      <validationMessage/>
+      <validationRegExp/>
+      <classType>com.xpn.xwiki.objects.classes.NumberClass</classType>
+    </size>
+  </class>
```
mflorea, line 133: Missing default class sheet binding (to show the list of properties when accessing the class page).

src/main/resources/XWiki/Macros/Translations.ro.xml
```diff
@@ -0,0 +1,25 @@
...
+  <title>Translations</title>
+  <comment/>
+  <minorEdit>true</minorEdit>
+  <syntaxId>plain/1.0</syntaxId>
+  <hidden>false</hidden>
+  <content>## Team macro&#xd;
+rendering.macro.team.content.options=Opțiune&#xd;
+rendering.macro.team.content.filter=A filtra…&#xd;
+rendering.macro.team.content.showUsernames=A arăta nume&#xd;
+rendering.macro.team.content.noUsers=Nu este pe nimeni să arate.</content>
```
mflorea, line 24:
```suggestion
rendering.macro.team.content.options=Opțiuni&#xd;
rendering.macro.team.content.filter=Filtrează…&#xd;
rendering.macro.team.content.showUsernames=Arată numele&#xd;
rendering.macro.team.content.noUsers=N-a fost găsit niciun utilizator.</content>
```

src/main/resources/XWiki/Macros/Translations.xml
```diff
@@ -71,7 +71,22 @@ rendering.macro.status.parameter.colour.description=The color of the status(text
 rendering.macro.status.parameter.title.name=Title
 rendering.macro.status.parameter.title.description=The text of the status. If not specified, the color name will be displayed (e.g.:"Grey").
 rendering.macro.status.parameter.subtle.name=Subtle
-rendering.macro.status.parameter.subtle.description=The aspect of the status: either white text on dark color background or dark color text on light color background. This is a boolean parameter so it should be filled with "true" or "false". Default: "false" (white text on dark color background).</content>
+rendering.macro.status.parameter.subtle.description=The aspect of the status: either white text on dark color background or dark color text on light color background. This is a boolean parameter so it should be filled with "true" or "false". Default: "false" (white text on dark color background).
+
+## Team macro
+rendering.macro.team.name=Team
+rendering.macro.team.name=description=Display avatars of the members of a team
+rendering.macro.team.parameter.tag.description=Tag that users need to have to be shown
```
mflorea, line 79: The parameter name also needs to be translated so it also needs to appear in the translation bundle.

src/main/resources/XWiki/Macros/Translations.xml
```diff
@@ -71,7 +71,22 @@ rendering.macro.status.parameter.colour.description=The color of the status(text
...
+rendering.macro.team.parameter.size.description=The size (height and width) of an avatar
+rendering.macro.team.parameter.showUsernames.description=The name of users should be displayed under their avatars
+rendering.macro.team.parameter.letterAvatarBgColor.description=The background color of the letter avatars
+rendering.macro.team.parameter.letterAvatarFontColor.description=The font color of the letter avatars
+rendering.macro.team.parameter.disableLetterAvatars.description=Don't use letter avatars and display a default avatar instead
+rendering.macro.team.parameter.requireExternalAuth.description=The user should be searched in LDAP or OAuth
+rendering.macro.team.content.options=Options
+rendering.macro.team.content.filter=Filter…
+rendering.macro.team.content.showUsernames=Show names
+rendering.macro.team.content.noUsers=There is nobody to show.</content>
```
mflorea, line 89: Missing translation key for administration section title and description. See https://github.com/xwikisas/application-licensing/blob/master/application-licensing-licensor/application-licensing-licensor-ui/src/main/resources/Licenses/Code/Translations.xml#L39-L40 .

src/main/resources/XWiki/Macros/TeamConfiguration.xml
```diff
@@ -0,0 +1,420 @@
+<?xml version='1.1' encoding='UTF-8'?>
+<xwikidoc version="1.4" reference="XWiki.Macros.TeamConfiguration" locale="">
```
mflorea, line 2: You need to make this page editable in the POM, because by default all pages that are part of a paid (licensed) extension are protected. Otherwise the users won't be able to use the administration section. See https://github.com/xwikisas/application-diagram/blob/master/application-diagram-ui/pom.xml#L43-L47 .

mflorea (changes requested):

raphj:
Thanks for the review, I'll apply these changes. Thank you for telling me about the mvn xar:format script, this is going to same me a lot of time.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+  <minorEdit>true</minorEdit>
+  <syntaxId>xwiki/2.0</syntaxId>
+  <hidden>false</hidden>
+  <content>{{velocity}}
+$services.localization.render('rendering.macro.team.description')
+
+== Parameters ==
+
+|=Name|=Description|=Default
+|tag|$services.localization.render('rendering.macro.team.parameter.tag.description')|projectteam
```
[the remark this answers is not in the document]
raphj: This is a leftover

```suggestion
|tag|$services.localization.render('rendering.macro.team.parameter.tag.description')|
```

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+      </use>
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
+      <code>(function () {
+  function filter (teamDiv, e) {
+    const l = this.value.toLowerCase()
+    for (const user of teamDiv.getElementsByClassName('xwikiteam-user')) {
```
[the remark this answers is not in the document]
raphj: No problem, I'll convert to a regular loop :-)
(for what it's worth, `for..of` on a `DOMList` breaks Safari 9 too)

src/main/resources/XWiki/Macros/Team.xml
raphj: You are right, I'll update this label

```suggestion
      <name>Team</name>
```

src/main/resources/XWiki/Macros/Team.xml
raphj: The default values are hard-coded, but I can move them out of the code to here if it's better.

src/main/resources/XWiki/Macros/Team.xml
raphj: That's something I didn't change rather than a conscious decision, I'll disable this.
```suggestion
      <supportsInlineMode>0</supportsInlineMode>
```

src/main/resources/XWiki/Macros/TeamConfiguration.xml
raphj:
```suggestion
  <hidden>true</hidden>
```

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+  <author>XWiki.RaphaelJakse</author>
+  <contentAuthor>XWiki.RaphaelJakse</contentAuthor>
+  <date>1647533963000</date>
+  <contentUpdateDate>1647531431000</contentUpdateDate>
+  <version>3.64</version>
+  <title/>
+  <comment/>
+  <minorEdit>true</minorEdit>
+  <syntaxId>xwiki/2.0</syntaxId>
+  <hidden>false</hidden>
```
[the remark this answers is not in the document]
raphj:
```suggestion
  <hidden>true</hidden>
```

src/main/resources/XWiki/Macros/Translations.xml
```diff
@@ -71,7 +71,22 @@ rendering.macro.status.parameter.colour.description=The color of the status(text
 rendering.macro.status.parameter.title.name=Title
 rendering.macro.status.parameter.title.description=The text of the status. If not specified, the color name will be displayed (e.g.:"Grey").
 rendering.macro.status.parameter.subtle.name=Subtle
-rendering.macro.status.parameter.subtle.description=The aspect of the status: either white text on dark color background or dark color text on light color background. This is a boolean parameter so it should be filled with "true" or "false". Default: "false" (white text on dark color background).</content>
+rendering.macro.status.parameter.subtle.description=The aspect of the status: either white text on dark color background or dark color text on light color background. This is a boolean parameter so it should be filled with "true" or "false". Default: "false" (white text on dark color background).
+
+## Team macro
+rendering.macro.team.name=Team
+rendering.macro.team.name=description=Display avatars of the members of a team
```
raphj, line 78: There is a equal to much here I think, I will this this.

src/main/resources/XWiki/Macros/TeamConfigurationClass.xml
```diff
@@ -0,0 +1,134 @@
...
+  <author>XWiki.RaphaelJakse</author>
+  <contentAuthor>XWiki.RaphaelJakse</contentAuthor>
+  <date>1647533770000</date>
+  <contentUpdateDate>1647417124000</contentUpdateDate>
+  <version>1.39</version>
+  <title/>
+  <comment>Renamed from xwiki:Macros.XWikiTeamMacro.ConfigurationClass</comment>
+  <minorEdit>true</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>false</hidden>
```
raphj, line 20: Should this be hidden too?

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+        obj2.className = 'XWiki.OIDC.UserClass'
+      )
+    #end and
+    obj.id = prop.id.id and
+    prop.id.name = 'active' and
+    prop.value = 1
+    $tagwhere
+  ORDER BY doc.name
+")
+#set($list = $services.query.hql($sql).execute())
```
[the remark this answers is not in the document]
raphj: Any guideline on the limit I should set and how? Would you have some time to discuss this?

src/main/resources/XWiki/Macros/Team.xml
raphj: Should  $pictureURL be escaped? (I realize that the other values need to be escaped)

src/main/resources/XWiki/Macros/Team.xml
raphj: How can I resolve the local user reference? Is there a way to do this in the hql query directly?

src/main/resources/XWiki/Macros/Team.xml
michitux: Yes, definitely. From a short look this may be a URL that may contain several parameters which are separated by `&` which needs to be escaped  - we try to produce valid XML and not rely on the fact that these are most likely not [ambiguous ampersands](https://mathiasbynens.be/notes/ambiguous-ampersands). There might also be other reasons (user-supplied attachment names we cannot trust, ...) but the fact that this might be a URL with parameters is enough.

raphj: ok, thanks!

src/main/resources/XWiki/Macros/Translations.xml
raphj: It seems done for the panel macro but not for the status macro, from which I took example of. Should we fix this?

src/main/resources/XWiki/Macros/Translations.xml
raphj: I'm lost here. When I grep that repository, I see no use of `admin.licenses.description`. What is the naming convention and how can I check it works?
Is the goal to translate <heading>Team Macro</head> in the configuration page and if so, how do I instruct XWiki to use that translation key there?

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+#macro(avatarInitials $name $size $letterAvatarBgColor $letterAvatarFontColor)
+  &lt;span class="xwikiteam-avatar xwikiteam-avatar-initials" style="height:${size}px; width:${size}px; border-radius: ${size}px; background-color: $letterAvatarBgColor"&gt;
+    &lt;span class="xwikiteam-avatar-initials-inner"&gt;
+      &lt;span class="xwikiteam-avatar-initials-letters" style="color: $letterAvatarFontColor"&gt;
+        #foreach($firstLetter in $name.toUpperCase().split("[\s-]"))$firstLetter.charAt(0)#end
+      &lt;/span&gt;
+    &lt;/span&gt;
+  &lt;/span&gt;
+#end
+#macro(avatar $username $size $disableLetterAvatars $letterAvatarBgColor $letterAvatarFontColor)
```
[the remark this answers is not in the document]
raphj: Ah, nice! That's way better now. It seems like resizedUserAvatar is an evolved version of part of the code I reworked.
I needed to use getUserAvatarURL to support initials avatars but maybe the initials avatar could be ported to the product itself if people like this.

src/main/resources/XWiki/Macros/Team.xml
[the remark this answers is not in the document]
raphj: I've added a limit parameter with a default value of 100 to fix this.

src/main/resources/XWiki/Macros/Team.xml
raphj: Ok, this works without putting "xwiki:". I guess the user will be correctly resolved without.

src/main/resources/XWiki/Macros/Team.xml
raphj: I did in the end.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1407 @@
...
+    <property>
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <code>{{velocity}}
+#set($discard = $xwiki.ssx.use('xwiki:XWiki.Macros.Team'))
+#set($discard = $xwiki.jsx.use('xwiki:XWiki.Macros.Team'))
+#set($pictureList = {})
+#macro(avatarInitials $name $size $letterAvatarBgColor $letterAvatarFontColor)
+  &lt;span class="xwikiteam-avatar xwikiteam-avatar-initials" style="height:${size}px; width:${size}px; border-radius: ${size}px; background-color: $letterAvatarBgColor"&gt;
```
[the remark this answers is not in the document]
mflorea: I don't understand the question. The values need to be escaped depending on how they are used. The escaping needs to be done inside the macro code, because we can't trust the values passed by the user as macro parameters.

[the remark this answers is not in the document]
raphj: avatarInitials is an internal macro that is called with value that have already been escaped. I guess eventually having avatarInitials as an autonomous macro would be cool so escaping inside the macro would be worth it, however I need to avoid double escaping.

I guess I can do this.

src/main/resources/XWiki/Macros/Team.xml
mflorea: We need to know whether the macro is supposed to list users from the current wiki or from the main wiki. I don't have the answer. I think it makes more sense to list the users from the current wiki (by default at least), but listing the users from the main wiki (even when you are not on the main wiki) is also a valid use case.

For listing the users from the current wiki, removing the "xwiki:" is fine because ``#avatar`` is probably resolving (directly or indirectly) the given user reference relative to the current wiki.

src/main/resources/XWiki/Macros/Translations.xml
mflorea: Yes, but with a separate PR.

src/main/resources/XWiki/Macros/Team.xml
raphj: That could be a parameter?
I'd suggest we resolve the users in the current wiki and add the parameter to resolve globally when needed/requested?

src/main/resources/XWiki/Macros/Translations.xml
raphj: https://github.com/xwikisas/xwiki-pro-macros/issues/65

src/main/resources/XWiki/Macros/Translations.xml
mflorea: You need to add:

```
admin.<sectionName>=Translated Section Name
admin.<sectionName>.description=Translated section descrition 
```

The key is computed by the Administration Application. You don't use this key yourself, you just have to provide the key that the Administration Application is looking for. And ``<sectionName>`` is what you specify in your configurable class object that defines the administration section.

src/main/resources/XWiki/Macros/Translations.fr.xml
```diff
@@ -0,0 +1,25 @@
...
+  <author>XWiki.RaphaelJakse</author>
+  <contentAuthor>XWiki.RaphaelJakse</contentAuthor>
+  <date>1647533271000</date>
+  <contentUpdateDate>1647533271000</contentUpdateDate>
+  <version>1.2</version>
+  <title>Translations</title>
+  <comment/>
+  <minorEdit>true</minorEdit>
+  <syntaxId>plain/1.0</syntaxId>
+  <hidden>false</hidden>
```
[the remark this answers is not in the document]
mflorea: Yes, the fact that a wiki page is used to localize the macro / application messages is an implementation detail that the user shouldn't be aware of.

src/main/resources/XWiki/Macros/TeamConfigurationClass.xml
mflorea: Definitely.

raphj:
```suggestion
  <hidden>true</hidden>
```

src/main/resources/XWiki/Macros/Translations.fr.xml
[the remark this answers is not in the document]
raphj:
```suggestion
  <hidden>true</hidden>
```

src/main/resources/XWiki/Macros/Translations.xml
raphj: So that would be "Pro Macros" ? Should I rename this to something like "promacros" and put "Pro Macros" in the Translation page like this?

```suggestion
rendering.macro.team.content.noUsers=There is nobody to show.</content>
admin.promacros=Pro Macros
admin.promacros.description=Settings for Pro Macros
```

src/main/resources/XWiki/Macros/TeamConfiguration.xml
```diff
@@ -0,0 +1,455 @@
...
+      <configurationClass>XWiki.Macros.TeamConfigurationClass</configurationClass>
+    </property>
+    <property>
+      <displayBeforeCategory/>
+    </property>
+    <property>
+      <displayInCategory>other</displayInCategory>
+    </property>
+    <property>
+      <displayInSection>Pro Macros</displayInSection>
```
raphj, line 278:
```suggestion
      <displayInSection>Pro Macros</displayInSection>
```

raphj, line 278:
```suggestion
      <displayInSection>promacros</displayInSection>
```

src/main/resources/XWiki/Macros/Team.xml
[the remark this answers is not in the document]
raphj: I did and there was no issues with double escaping in the end. Values passed to this macro that risk double escaping have no legal values where double escaping would pose a problem.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1491 @@
...
+#set($disableTools = $wikimacro.parameters.disableTools)
+#set($disableLetterAvatars = $wikimacro.parameters.disableLetterAvatars)
+#set($limit = $wikimacro.parameters.limit)
+#set($configObject = $xwiki.getDocument("XWiki.Macros.TeamConfiguration").getObject("XWiki.Macros.TeamConfigurationClass"))
+#if("$!{size}" == "")
+  #set($size = $configObject.getValue("size"))
+#end
+#set($size = $escapetool.xml($size))
+#if("$!{size}" == "")
+  #set($size = 60)
```
mflorea, line 717: To be checked, but I think this line (as well as those below) is not reached now that you have the default value specified in the macro parameter descriptor. It's not a problem for the size parameter, but fort the rest, if you need to rely on external configuration (i.e. control de default values from the administration) it will be a problem because the value configured in the administration is not going to be used. This approach is not common and in fact this is the first macro I know that comes with a configuration section to control the default parameter values. The problem is that there's no perfect solution ATM, I think:

* the parameter descriptor should specify the default values so that the WYSIWYG editor doesn't output ``param=value`` when inserting a macro and using the default parameter value
* you need the default values outside of macro definition which is read-only for licensed extensions, if you want to allow the user to change them

This deserves a discussion on the forum to see what is the best way to make the default macro parameter values configurable.

mflorea (changes requested):

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1488 @@
...
+
+.xwikiteam .xwikiteam-user a {
+  display:flex;
+  flex-direction: column;
+  align-items: center;
+  margin: 0.4ex;
+}
+
+.xwikiteam .xwikiteam-avatar {
+  border: 4px solid white;
```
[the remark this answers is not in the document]
raphj: It should thanks to the box shadow

image
image

There might be situation I didn't think of however.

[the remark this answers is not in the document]
mflorea: OK, we'll see if we get complains.

src/main/resources/XWiki/Macros/Team.xml
raphj: I'd argue for:

mflorea: The problem is that if you define the default values outside of the macro parameter descriptor then anyone relying on the macro parameter descriptor (e.g. the WYSIWYG editor) to read the default values will get the wrong information. E.g. if some tool generates documentation for macros and shows the default values for macro parameters, then that automatically generated documentation will be wrong for the teams macro. I still think this needs a discussion.

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1488 @@
...
+    if (filterInput) {
+      filterInput.addEventListener('keyup', filter.bind(filterInput, teamDiv));
+      if (filterInput.value) {
+        filter.call(filterInput, teamDiv);
+        toggleTools(teamDiv)
+      }
+    }
+
+    bindShowUsernameCheckbox(teamDiv, teamDiv.querySelector(".xwikiteam-show-username-checkbox"));
+  }
```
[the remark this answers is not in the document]
raphj: done!

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1501 @@
...
+          filter.call(filterInput, teamDiv);
+          toggleTools(teamDiv)
+        }
+      }
+
+      bindShowUsernameCheckbox(teamDiv, teamDiv.querySelector(".xwikiteam-show-username-checkbox"));
+    }
+  }
+
+  $(document).on('xwiki:dom:updated', init);
```
mflorea, line 250: Where is ``$`` defined? You simply copied the code, but ``$`` was actually a parameter of the require block (that you don't have), and was pointing to jQuery. In your code ``$`` currently points to ``Prototype.js``.

raphj: Whoops, I expected JQuery to be unconditionally available in XWiki. I'll rewrite this in classic javascript.

raphj: fixed

raphj: Note that I haven't managed to make this work in the wysiwyg editor, but I don't think the features enabled by this script are relevant in the editor. It's probably better if they are not there when editing.

raphj: > You simply copied the code

This is extra confusing because the code ran as expected when testing (on page load)

src/main/resources/XWiki/Macros/Team.xml
raphj: https://forum.xwiki.org/t/handling-configurable-default-parameters-in-a-macro/10094

src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -0,0 +1,1505 @@
...
+          filter.call(filterInput, teamDiv);
+          toggleTools(teamDiv)
+        }
+      }
+
+      bindShowUsernameCheckbox(teamDiv, teamDiv.querySelector(".xwikiteam-show-username-checkbox"));
+    }
+  }
+
+  document.addEventListener('xwiki:dom:updated', init);
```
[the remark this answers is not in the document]
raphj: So, it seems like XWiki indeed triggers `xwiki:dom:updated` either using JQuery, either `document.fire`, which is a [Prototype JS function](https://github.com/prototypejs/prototype/blob/master/src/prototype/dom/event.js#L845).

`document.fire` seems to use [document.createEvent](https://developer.mozilla.org/en-US/docs/Web/API/Document/createEvent), [Event.initEvent](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent) and [EventTarget.dispatchEvent](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent), so that should work.

However, it seems it does not work if JQuery is used to trigger the event: only one alert is show with the following code:

```html
<!DOCTYPE html>
<script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
<script>
  addEventListener("DOMContentLoaded", function () {
    document.addEventListener("testcustomevent", function () {
      alert("triggered natively");
    })

    $(document).on("testcustomevent", function () {
      alert("triggered from JQuery");
    })

    $(document).trigger("testcustomevent", {})
  })
</script>
```
Anyway, the [XWiki documentation](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/JavaScriptAPI/) advises to use `document.observe`

Now, I can notice that the xwiki:dom:updated event is indeed triggered and that the initTools function from my last commit is called. However, when editing the macro with the WYSIWYG editor, I noticed that the script does not find the `xwikiteam` elements, and that seems to be because the script is run in the context of the outer window, not from the iframe editor.

Is this a bug or a feature (of the platform)?

Anyway, for this macro, it's fine for me that the tools don't appear in the editor (for now) and it would be fine if they would at some point. The tools are intended for viewing. They seem rather useless for edition but they would not hurt neither.

src/main/resources/XWiki/Macros/Team.xml
raphj: @mflorea would using advanced parameters as suggested by Vincent here https://forum.xwiki.org/t/handling-configurable-default-parameters-in-a-macro/10094/2 be an acceptable solution? It requires using Java code. Or should we go with something simpler and make this macro less configurable by not allowing users to use parameters or admins to set defaults? WDYT?

raphj: I decided to remove the configuration class for now, we can always make the team macro configurable in the future if needed

pom.xml
```diff
@@ -48,6 +48,54 @@
...
+      <artifactId>macro-jwplayer</artifactId>
+      <version>${jwplayer.version}</version>
+      <type>xar</type>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-rendering-wikimacro-store</artifactId>
+      <version>${platform.version}</version>
+    </dependency>
+  </dependencies>
```
raphj, line 98:
```suggestion
```

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Translations.xml
```diff
@@ -124,6 +150,8 @@ rendering.macro.microsoftStream.parameter.autoplay.description=Whether the video
         <cache>0</cache>
         <disabled>0</disabled>
         <displayType>select</displayType>
+        <freeText>forbidden</freeText>
+        <largeStorage>0</largeStorage>
```
raphj, line 154:
```suggestion
```

src/main/resources/XWiki/Macros/Team.xml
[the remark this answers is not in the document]
mflorea: > Anyway, the [XWiki documentation](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/JavaScriptAPI/) advises to use document.observe

That part of the documentation is outdated. You shouldn't use Prototype.js anymore. You should use jQuery for listening to **custom** XWiki events such as ``xwiki:dom:updated``. For standard events plain JavaScript is perfectly fine.

> However, when editing the macro with the WYSIWYG editor, I noticed that the script does not find the xwikiteam elements, and that seems to be because the script is run in the context of the outer window, not from the iframe editor.

The JavaScript code is executed in the context of the Window that loads it:
* for in-place WYSIWYG editing there's no inner frame / window, so all the JavaScript code is loaded in the main / top level window, and it should be able to find the ``xwikiteam`` elements within the edited content
* for stand-alone WYSIWYG editing the edited content is within an iframe, so nested Window; in this case it depends who loads the JavaScript code; if it's the edited content, then the JavaScript code should be loaded inside the nested Window. But not that JavaScript execution is "disabled" by default in the nested content Window by default, see https://extensions.xwiki.org/xwiki/bin/view/Extension/CKEditor+Integration#HExecuteJavaScriptcodeinsidetheeditingarea28XWiki10.102B29 .

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -221,33 +219,51 @@ $services.localization.render('rendering.macro.team.description')
...
+        }
+      }
+
+      bindShowUsernameCheckbox(teamDiv, teamDiv.querySelector(".xwikiteam-show-username-checkbox"));
     }
+  }
+
+  function init() {
+    initTools();
+    document.observe('xwiki:dom:updated', initTools);
```
mflorea, line 252: Don't use Prototype.js . Use jQuery for listening to custom XWiki events (as per https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/FrontendResources/JavaScriptAPI/#HBridgingcustomXWikieventsbetweenPrototypeandjQuery ) Also, when ``xwiki:dom:updated`` is triggered you should re-initialize **only** the ``xwikiteam`` elements that have been updated.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -221,33 +219,51 @@ $services.localization.render('rendering.macro.team.description')
...
+  }
 
-    bindShowUsernameCheckbox(teamDiv, teamDiv.querySelector(".xwikiteam-show-username-checkbox"));
+  if (document.body) {
+    init();
+  } else {
+    window.addEventListener('DOMContentLoaded', init);
   }
```
mflorea, line 259: jQuery can make this simpler. See for instance https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-war/src/main/webapp/resources/uicomponents/suggest/xwiki.selectize.js#L333-L334 .

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -656,50 +674,46 @@ $services.localization.render('rendering.macro.team.description')
...
-          #set($pictureURL = $xwiki.getAttachmentURL('XWiki.XWikiUserSheet', 'noavatar.png'))
-        #end
-        #set($picture = "&lt;img class='xwikiteam-avatar' src='$pictureURL' alt='$displayUser' title='$displayUser' width='$size' height='$size' /&gt;")
+        &lt;img
+          class='xwikiteam-avatar'
+          src='$escapetool.xml($return.url)'
+          alt='$escapedDisplayUser'
+          title='$escapedDisplayUser'
+          width='$size'
+          height='$size'
```
mflorea, line 713: Is the size (XML) safe? I see it is taken from configuration, so it could be anything. Why not escape its value?

raphj, line 713:
```suggestion
          width='$escapetool.xml($size)'
          height='$escapetool.xml($size)'
```

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -221,33 +219,51 @@ $services.localization.render('rendering.macro.team.description')
...
+  function initTools() {
+    const teamDivs = document.getElementsByClassName('xwikiteam');
+    for (let i = 0, len = teamDivs.length; i &lt; len; i++) {
+      const teamDiv = teamDivs[i];
+      const a = teamDiv.querySelector('.xwikiteam-tools-link a');
+      if (a) {
+        if (a.onclick) {
+          continue; // don't initialize already initialized team macros
+        }
+        a.onclick = toggleTools.bind(null, teamDiv);
```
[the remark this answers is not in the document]
raphj:
```suggestion
        a.addEventListener('click', toggleTools.bind(null, teamDiv));
```

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
raphj: Fixed, I escaped the values.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
raphj: Fixed.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
raphj: fixed

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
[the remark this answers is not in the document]
raphj: Okay, I now use a class name to now if the event handler has already been bound to the link

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -26,32 +26,30 @@
...
+  <version>1.1</version>
   <title/>
   <comment/>
-  <minorEdit>true</minorEdit>
-  <syntaxId>xwiki/2.0</syntaxId>
-  <hidden>false</hidden>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>true</hidden>
   <content>{{velocity}}
```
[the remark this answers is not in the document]
raphj: No indeed!

src/main/resources/XWiki/Macros/Team.xml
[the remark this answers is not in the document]
raphj: Ok, thanks for the information.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -247,17 +246,11 @@
...
-    document.observe('xwiki:dom:updated', initTools);
-  }
-
-  if (document.body) {
-    init();
-  } else {
-    window.addEventListener('DOMContentLoaded', init);
-  }
-})();</code>
+    $('xwiki:dom:updated', initTools);
```
mflorea, line 251: You're still iterating all the ``xwikiteam`` elements whenever ``xwiki:dom:updated`` is triggered (instead of looking **just** at the updated elements, which you can get from the event data), but you're initializing only the ``xwikiteam`` elements that are not already initialized, which I guess it's fine. My worry is that the ``initialized`` CSS class is not bulletproof. Take this use case:

* you are inside the WYSIWYG editor
* the found ``xwikiteam`` elements are initialized, thus getting the ``initialized`` CSS class
* the editor saves the current HTML (in order to be able to Undo / Redo)
* the user makes some changes and then clicks on Undo
* the editor reverts the HTML and triggers the ``xwiki:dom:updated`` event
* the ``xwikiteam`` elements are **not** re-initialized because they have the ``initialized`` CSS class

I haven't tested but this user case make sense. All in all, I think it's safer to re-initialize everything from the updated content indicated by ``xwiki:dom:updated``.

raphj: Fixed by removing the use of a class and removing / adding the event unconditionally.

PR #67, opened 2022-05-09
Develop the Tip macro

### LINKED ISSUE #66: Develop the Tip macro
This is nearly the same as XWikis `success` macro.

ClemensRobbenhaar:
Fixes #66


src/main/resources/Confluence/Macros/Tip.xml
```diff
@@ -0,0 +1,370 @@
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <code>{{velocity}}
+  {{success}}
+    #if("$!wikimacro.parameters.title" != "")
+    **$!{wikimacro.parameters.title}**
+
+    #end
+    $wikimacro.content
```
mflorea, line 260: This means we won't be able to use Velocity macro inside the Tip macro content, unless you wrap the Tip macro with the Velocity macro. Maybe that's fine, just wan to make sure you are aware.

mflorea: Couldn't you use ``{{wikimacrocontent/}}``?

ClemensRobbenhaar: About the velocity macro: I hope that content migrated from Confluence does not have velocity macros embedded in their "infoboxes" macros. Then maybe instead of writing these wrapper macros we might better add them to the macro-mapping in the Confluence Migrator (if that macro mapping has been implemented, I forgot to check the current state, sorry.) That might avoid using velocity at all.

ClemensRobbenhaar: I have updated the PR :  the "tip" marco (also also the "note" macro) now uses the ``{{wikimacrocontent/}}``, and also ``{{wikimacroparameter name="title" /}}``.

The latter inserts an empty line on top of the title, however.
Old version without ``wikimacroparameter``:
tip-nowikiparam

New version with ``wikimacroparameter``:
tip-wikiparam

I think I can hush this up via CSS if necessary, but that might be ugly :-/

Issue #74, opened 2022-08-03
Create a user list macro


slauriere:
This macro will make it possible to select users and to display them in a list with their avatar and a link to their profile.

In distinct tasks, the macro could consider compatibility with existing Confluence macros: 
* [Confluence User List macro](https://confluence.atlassian.com/doc/user-list-macro-139546.html)
* Linchpin Confluence Customized User List macro

PR #75, opened 2022-08-03
#74: Create a user list macro

slauriere:
This PR implements a user list macro. The code is strongly inspired by the one of the PDFViewer macro. Here's the below a screen capture of the macro in edit mode and rendered in the wysiwyg editor. Review and feedback welcome.

userlist

Note that there's currently a known issue when editing the user list that is very similar to xwikisas/macro-pdfviewer#43.


xwiki-pro-macros-api/src/main/resources/templates/userlist/userlist.vm
```diff
@@ -0,0 +1,17 @@
+#set ($users = $params.users)
+#set ($users = $users.split(','))
+#set ($discard = $xwiki.ssx.use('XWiki.Macros.UserListMacroStyleSheet'))
+<ul class="xwiki-userlist">
+#foreach ($user in $users)
```
mflorea, line 5: We already have a Velocity macro to display a list of users. See https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm#L246 .

    @param arg could be either a string, a reference or a list of them

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
```diff
@@ -0,0 +1,86 @@
...
+    {
+        super("User list", "Displays a custom list of users with their avatar",
+            UserListMacroParameters.class);
+    }
+
+    @Override
+    public List<Block> execute(UserListMacroParameters parameters, String content, MacroTransformationContext context)
+        throws MacroExecutionException
+    {
+        Template customTemplate = this.templateManager.getTemplate("userlist/userlist.vm");
```
mflorea, line 64: Why not use ``/html_displayer/userresourcereference/view.vm`` because a displayer supports both **edit** and **view**, and in your case you basically need a picker (edit) for a list of users and a displayer (view) for it.

Moreover, you should actually use the displayer API to render the user list in view mode. Checkout what ``$services.display.html.display(...)`` does under the hood (used in https://github.com/xwiki-contrib/application-ckeditor/blob/master/application-ckeditor-ui/src/main/resources/CKEditor/MacroService.xml#L280 by CKEditor to get the pickers for the macro parameters).

But in order to do this it would be more accurate to use ``List<UserResourceReference>`` as display type for the macro parameter (so that you can pass a list of values when rendering in view mode).

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/UserResourceReference.java
```diff
@@ -0,0 +1,43 @@
...
+import org.xwiki.rendering.listener.reference.ResourceReference;
+import org.xwiki.rendering.listener.reference.ResourceType;
+
+/**
+ * The object is used to associate a user picker to the reference parameter.
+ *
+ * @version $Id$
+ * @since 1.0
+ */
+public class UserResourceReference extends ResourceReference
```
mflorea, line 31: Why ``ResourceReference`` and not ``DocumentReference``? You mention below:

> A user is represented by a using a serialized DocumentReference.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
```diff
@@ -0,0 +1,50 @@
...
+    public String getUsers()
+    {
+        return this.users;
+    }
+
+    /**
+     * Sets a list of users. A user is represented by a using a serialized DocumentReference.
+     * @param users list of user references
+     */
+    @PropertyDisplayType(UserResourceReference.class)
```
mflorea, line 44: It would be more accurate to use a ``List`` type, or something that extends a list. Bu you probably need to create a ``UserResourceReferenceList`` class. The advantage is that you'll be able to pass a list value when displaying in view mode (see my previous comment).

xwiki-pro-macros-api/src/main/resources/templates/userlist/userlist.vm
```diff
@@ -0,0 +1,17 @@
+#set ($users = $params.users)
+#set ($users = $users.split(','))
+#set ($discard = $xwiki.ssx.use('XWiki.Macros.UserListMacroStyleSheet'))
```
mflorea, line 3: This makes the API module depend on the UI (which depends on the API). Would be better to have the CSS either in the API module (``ssrx``) or in a new WebJar module.

xwiki-pro-macros-api/pom.xml
```diff
@@ -49,5 +49,12 @@
       <groupId>org.webjars</groupId>
       <artifactId>jquery</artifactId>
     </dependency>
+    <!-- Dependencies needed by the User List macro -->
+    <!-- Needed to use the default TemplateManager -->
```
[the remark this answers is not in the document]
slauriere: Right, thanks. I updated this dependency in the latest commit.

xwiki-pro-macros-api/src/main/resources/templates/userlist/userlist.vm
slauriere: Indeed, I had not noticed that one but the list of user properties to be displayed will have to be configurable from the macro so I'm not sure we will be able to reuse this existing `displayUser` macro.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
slauriere: The commit d4f70db3 introduces a `UserReferenceList` class used as a macro parameter indeed. It also brings together the edit and view templates in a single common folder, renamed to `userreferencelist` to match the new parameter type.

I will see what I can do with the displayer API.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
```diff
@@ -0,0 +1,91 @@
...
+        throws MacroExecutionException
+    {
+        Template customTemplate = this.templateManager.getTemplate("html_displayer/userreferencelist/view.vm");
+
+        try {
+            this.bindParameters(parameters);
+            return this.templateManager.execute(customTemplate).getChildren();
+        } catch (Exception e) {
+            throw new MacroExecutionException("Failed to render the userList viewer template.", e);
+        }
```
mflorea, line 77: The path to the HTML displayer template, it's name and the way the parameters are bind is an implementation detail of the HTML displayer. Your code shouldn't be aware of that (i.e. you shouldn't have to update this code if the HTML displayer implementation changes). Why not use directly the HTML displayer API? See https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-html/src/main/java/org/xwiki/rendering/display/html/script/HTMLDisplayerScriptService.java .

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserReferenceListConverter.java
```diff
@@ -0,0 +1,67 @@
...
+import org.xwiki.properties.converter.ConversionException;
+
+import com.xwiki.macros.userlist.macro.UserReferenceList;
+
+/**
+ * XWiki Properties Bean Converter to convert Strings into {@link UserReferenceList}.
+ *
+ * @see org.xwiki.properties.converter.Converter
+ *
+ *  @version $Id$
```
mflorea, line 39: Strange formatting.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
```diff
@@ -0,0 +1,74 @@
...
+    {
+        return this.users;
+    }
+
+    /**
+     * Sets a list of users.
+     *
+     * @param users list of user references
+     */
+    @PropertyDisplayType(UserReferenceList.class)
```
mflorea, line 48: Is this needed if the parameter type is ``UserReferenceList``? (it's redundant).

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
```diff
@@ -0,0 +1,74 @@
...
+package com.xwiki.macros.userlist.macro;
+
+import java.util.Arrays;
+import java.util.List;
+
+import org.xwiki.properties.annotation.PropertyDescription;
+import org.xwiki.properties.annotation.PropertyDisplayType;
+import org.xwiki.properties.annotation.PropertyName;
+
+public class UserListMacroParameters
```
mflorea, line 29: Missing JavaDoc.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserReferenceList.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.ArrayList;
+
+import org.xwiki.model.reference.EntityReference;
+
+/**
+ * Class representing a list of user references. This class is needed to declare a valid
+ * @{@link org.xwiki.properties.annotation.PropertyDisplayType} in @{@link UserListMacroParameters}.
+ *
+ * @version $Id$
+ *
```
mflorea, line 31: Empty line not needed.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
```diff
@@ -0,0 +1,50 @@
...
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+## Lesser General Public License for more details.
+##
+## You should have received a copy of the GNU Lesser General Public
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+## Hack copied from https://github.com/xwiki-contrib/application-changerequest/blob/main/application-changerequest-ui/src/main/resources/ChangeRequest/Code/SaveChangeRequestModal.xml#L42
```
mflorea, line 20: You should link to a tag, otherwise the linked code can change.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
```diff
@@ -0,0 +1,50 @@
...
+              #else
+                #set ($displayedProperty = $userObj.display($propertyName, 'view'))
+                #stripHTML($displayedProperty)
+              #end
+            </td>
+          #end
+        </tr>
+      #end
+    #end
+  #end
```
mflorea, line 49: I'm worried about the empty table element generated when there are no users. I don't think it's valid HTML. Maybe show a message in this case?

mflorea (changes requested):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
slauriere: Thanks for the explanation. I modified the code to use the `HTMLDisplayerManager`. Not sure this is completely in line with what you meant though since you mentioned the `HTMLDisplayerScriptService`.  What do you think?

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserReferenceListConverter.java
slauriere: I fixed the formatting.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
slauriere: I removed this redundancy.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
slauriere: I added a JavaDoc.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserReferenceList.java
slauriere: I removed this empty line.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
slauriere: I linked to a tag indeed.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
slauriere: I modified the code to make sure a table is generated only when there are no users, and displaying a warning message otherise. Also added a table row containing a warning message for invalid users.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
slauriere: Using this approach, I'm running into an issue with the avatar: the HTML for the avatar image is not generated correctly, partially interpreted as wiki syntax it seems, with the following result scheme (with the unexpected `wikigeneratedid` class):

```
<img src="/xwiki/bin/download/Sandbox/UserList/WebHome/%2Fxwiki%2Fbin%2Fskin%2Fresources%2Ficons%2Fxwiki%2Fnoavatar.png%3Fcache-version%3D1654607551492" alt="Bob" title="Bob" class="avatar avatar_50 wikigeneratedid" id="I2Fxwiki2Fbin2Fskin2Fresources2Ficons2Fxwiki2Fnoavatar.png3Fcache-version3D1654607551492">```
```

I could not understand why so far but maybe I'm not using the `HTMLDisplayer*` services along what you had in mind so I'll 
wait for your input about that before digging further.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
slauriere: Now using a `UserReferenceList` indeed, which extends `ArrayList<EntityReference>` (we may want to extend `ArrayList<UserReference>` in the future when upgrading the dependency on the platform to a more recent version).

xwiki-pro-macros-api/src/main/resources/templates/userlist/userlist.vm
slauriere: Now using a CSS via `ssrx` indeed.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/UserResourceReference.java
slauriere: Now using `EntityReference`, as a list (see comment below).

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
slauriere: Thank you for all the remarks and guidance, it's very instructive.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
slauriere: I fixed the avatar issue by generating a `new RawBlock(html, Syntax.HTML_5_0)` directly rather than using an html parser (the initial `@Named("xhtml/5")` Parser was wrong, but even when using a `html/5.0` I was not getting the expected blocks, I don't understand well why).

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
```diff
@@ -19,40 +19,47 @@
...
 @Singleton
 public class UserListMacro extends AbstractMacro<UserListMacroParameters>
 {
     @Inject
-    private TemplateManager templateManager;
+    private HTMLDisplayerManager htmlDisplayerManager;
 
     @Inject
-    private ScriptContextManager scriptContextManager;
+    @Named("default")
```
mflorea, line 61: You should just omit the ``@Named`` annotation when the hint is "default" (your code works, but it can be simplified).

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
```diff
@@ -67,22 +74,21 @@ public UserListMacro()
...
-            this.bindParameters(parameters);
-            return this.templateManager.execute(customTemplate).getChildren();
-        } catch (Exception e) {
-            throw new MacroExecutionException("Failed to render the userList viewer template.", e);
+            List<String> userReferences =
+                parameters.getUsers().stream().map(reference -> referenceSerializer.serialize(reference)).collect(
+                    Collectors.toList());
+            params.put("users", StringUtils.join(userReferences, ','));
+            params.put("properties", StringUtils.join(parameters.getProperties(), ','));
+            String html = htmlDisplayerManager.display(UserReferenceList.class, content, params, "view");
```
mflorea, line 85: The second parameter is the displayed **value**. Isn't the list of user references the displayed value? I don't understand why you pass the ``content`` which should be empty all the time since your macro doesn't support content. I think you should pass directly ``parameters.getUsers()``, and keep the third parameter (``params``) only for stuff that controls how the value (the list of user references) are displayed (i.e. the list of user properties).

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
```diff
@@ -17,34 +17,51 @@
...
-                #stripHTML($displayedProperty)
-              #end
-            </td>
-          #end
-        </tr>
+#macro (displayInvalidUserMessageRow $user $colspan)
+  <tr>
+    <td colspan="$colspan">
+      <div class="box warningmessage">
+        $services.localization.render('rendering.macro.userList.invalidUser', [$user])
```
mflorea, line 28: Missing XML escaping of the translation result.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
```diff
@@ -17,34 +17,51 @@
...
+      </div>
+    </td>
+  </tr>
+#end
+#if ("$!displayer.parameters.users" != '')
+  #set ($users = $displayer.parameters.users.split(','))
```
mflorea, line 34: You should use ``$displayer.value`` instead, because the user list is the displayed value. See my previous comment.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
```diff
@@ -17,34 +17,51 @@
...
+      #else
+        #displayInvalidUserMessageRow($user, $propertyNames.size())
       #end
     #end
-  #end
-</table>
\ No newline at end of file
+  </table>
+#else
+  <div class="box warningmessage">$services.localization.render('rendering.macro.userList.isEmpty')</div>
```
mflorea, line 66: Missing XML escape.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
```diff
@@ -17,34 +17,51 @@
...
+    </td>
+  </tr>
+#end
+#if ("$!displayer.parameters.users" != '')
+  #set ($users = $displayer.parameters.users.split(','))
+  #set ($propertyNames = $displayer.parameters.properties.split(','))
+  #set ($discard = $xwiki.ssrx.use('css/userlist.css'))
+  <table class="xwiki-userlist">
+    #foreach ($user in $users)
+      #if ($xwiki.exists($user) && $services.security.authorization.hasAccess('view', $xcontext.user, $user))
```
mflorea, line 39: You can omit the ``$xcontext.user`` because that's the default: checks access rights for the current user:

```suggestion
      #if ($xwiki.exists($user) && $services.security.authorization.hasAccess('view', $user))
```

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
```diff
@@ -17,34 +17,51 @@
...
+        #set ($userDoc = $xwiki.getDocument($user))
+        #set ($userObj = $userDoc.getObject('XWiki.XWikiUsers'))
+        #if ($userObj != $NULL)
+          <tr>
+            #foreach ($propertyName in $propertyNames)
+              <td class="xwiki-userlist-user-$propertyName">
```
mflorea, line 45: Missing XML escape for the property name (it comes from the person calling the macro, so not trusted).

mflorea (changes requested):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
slauriere: Right, done in latest commit

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
slauriere: You're right, the displayed value is the list of user references. I updated the code to pass this value instead of the content. Note that in edit mode, the displayer value is null when rendering `userreferencelist/edit.vm`, I don't understand why. This prevents from editing the user list correctly in the macro dialog. Would you know how to fix this?

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
slauriere: fixed in latest commit

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
slauriere: fixed indeed

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
slauriere: done

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreferencelist/view.vm
slauriere: I escaped it indeed.

mflorea (approved):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListMacro.java
mflorea: > Note that in edit mode, the displayer value is null when rendering userreferencelist/edit.vm, I don't understand why. This prevents from editing the user list correctly in the macro dialog.

The Edit Macro (parameters) dialog doesn't call the parameter displayer with the actual parameter value (mostly because when inserting a macro we don't have a value yet), but with the default value (from the parameter descriptor), which in your case is probably null. This way it obtains an edit "template" for the macro parameter, which is used both when inserting a new macro and when editing an existing one. See https://github.com/xwiki-contrib/application-ckeditor/blob/master/application-ckeditor-ui/src/main/resources/CKEditor/MacroService.xml#L280-L282 :

* the edit template is rendered on the server side using the default value
* the actual value is set on the client side

So what is exactly the problem? The user picker is not initialized (on the client side) with the macro parameter value when editing an existing macro? Either the edit (HTML) "template" returned from the server side is not right or the code that sets the value on the client side has a bug / limitation. But I think we have other macros that use a picker (e.g. office macro) and for which editing an existing macro works fine.

mflorea: I just tested and editing an existing office macro call works as expected (the value is properly set on the picker).

Issue #79, opened 2022-09-01
Create a user profile macro


slauriere:
This macro will allow to display a user profile in a card: user name, avatar and a customizable list of attributes.

* See also: [Confluence profile macro](https://confluence.atlassian.com/doc/user-profile-macro-189432828.html)

PR #80, opened 2022-09-05
#79: Create a User Profile macro

slauriere:
This pull request adds a UserProfile macro (id: `userProfile`) with its parameter class, custom displayer and translation entries. The list of displayed properties can be customized using the `properties` macro parameter. Here's below how it renders currently, with a user picker on edit. It is styled by the stylesheet `css/userprofile.css`. That'd be better to use a LESS stylesheet so as to use some of the current theme colors but I could not find how to do this in this context for now. I will try to investigate, any hint welcome.

userprofile

### DIFF
--- a/xwiki-pro-macros-api/pom.xml
+++ b/xwiki-pro-macros-api/pom.xml
@@ -49,5 +49,16 @@
       <groupId>org.webjars</groupId>
       <artifactId>jquery</artifactId>
     </dependency>
+    <!-- Dependencies needed by the User List macro -->
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-template-api</artifactId>
+      <version>${platform.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.commons</groupId>
+      <artifactId>xwiki-commons-displayer</artifactId>
+      <version>${commons.version}</version>
+    </dependency>
   </dependencies>
 </project>
--- a/xwiki-pro-macros-api/pom.xml
+++ b/xwiki-pro-macros-api/pom.xml
@@ -49,16 +49,26 @@
       <groupId>org.webjars</groupId>
       <artifactId>jquery</artifactId>
     </dependency>
-    <!-- Dependencies needed by the User List macro -->
-    <dependency>
-      <groupId>org.xwiki.platform</groupId>
-      <artifactId>xwiki-platform-template-api</artifactId>
-      <version>${platform.version}</version>
-    </dependency>
+    <!-- Dependencies needed by the User Profile macro -->
     <dependency>
       <groupId>org.xwiki.commons</groupId>
       <artifactId>xwiki-commons-displayer</artifactId>
       <version>${commons.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-api</artifactId>
+      <version>${rendering.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.rendering</groupId>
+      <artifactId>xwiki-rendering-transformation-macro</artifactId>
+      <version>${rendering.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-model</artifactId>
+      <version>${platform.version}</version>
     </dependency>
   </dependencies>
 </project>
--- a/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userprofile/internal/macro/UserProfileMacro.java
+++ b/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userprofile/internal/macro/UserProfileMacro.java
@@ -1,0 +1,94 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xwiki.macros.userprofile.internal.macro;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.displayer.HTMLDisplayerException;
+import org.xwiki.displayer.HTMLDisplayerManager;
+import org.xwiki.model.reference.EntityReferenceSerializer;
+import org.xwiki.rendering.block.Block;
+import org.xwiki.rendering.block.RawBlock;
+import org.xwiki.rendering.macro.AbstractMacro;
+import org.xwiki.rendering.macro.MacroExecutionException;
+import org.xwiki.rendering.syntax.Syntax;
+import org.xwiki.rendering.transformation.MacroTransformationContext;
+import org.xwiki.text.StringUtils;
+
+import com.xwiki.macros.userprofile.macro.UserProfileMacroParameters;
+import com.xwiki.macros.userprofile.macro.UserReference;
+
+/**
+ * This macro displays a user profile.
+ *
+ * @version $Id$
+ */
+@Component
+@Named("userProfile")
+@Singleton
+public class UserProfileMacro extends AbstractMacro<UserProfileMacroParameters>
+{
+    @Inject
+    private HTMLDisplayerManager htmlDisplayerManager;
+
+    @Inject
+    @Named("default")
+    private EntityReferenceSerializer<String> referenceSerializer;
+
+    /**
+     * Create and initialize the descriptor of the macro.
+     */
+    public UserProfileMacro()
+    {
+        super("User profile", "Displays a user profile with custom set of properties",
+            UserProfileMacroParameters.class);
+    }
+
+    @Override
+    public List<Block> execute(UserProfileMacroParameters parameters, String content,
+        MacroTransformationContext context) throws MacroExecutionException
+    {
+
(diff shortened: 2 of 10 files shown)


slauriere:
I updated `UserProfileMacroParameters` to declare a default set of properties however I noticed that this makes it impossible to reset this property set to an empty list from the interface, could not figure out why yet.

mflorea:
> It is styled by the stylesheet css/userprofile.css. That'd be better to use a LESS stylesheet so as to use some of the current theme colors but I could not find how to do this in this context for now. I will try to investigate, any hint welcome.

That's probably because currently ``ssrx`` doesn't support loading LESS files (i.e. doesn't support LESS compilation). I know that we support loading LESS files from WebJars (and compiling them in the context of the current skin), so one option is to add a WebJar module to the Pro Macros to hold client-side resources, but the issue is that this feature is relatively recent https://jira.xwiki.org/browse/XWIKI-18492 (13.3).

mflorea:
> I updated UserProfileMacroParameters to declare a default set of properties however I noticed that this makes it impossible to reset this property set to an empty list from the interface, could not figure out why yet.

The suggest input (based on Selectize) has code to cover precisely this use case, see https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-14.7/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm#L2732-L2736 . There's a hidden input with empty value added after the select, that is used to submit the empty selection when the select element has no option selected.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreference/view.vm
```diff
@@ -0,0 +1,78 @@
...
+#if ("$!ref" != '' && $xwiki.exists($reference) && $services.security.authorization.hasAccess('view', $reference))
+  #set ($discard = $xwiki.ssrx.use('css/userprofile.css'))
+  #set ($properties = $displayer.parameters.properties.split(','))
+  #set ($userPage = $xwiki.getDocument($reference))
+  #set ($userObj = $userPage.getObject('XWiki.XWikiUsers'))
+  <div class="xwiki-user-profile-box">
+    <div class="xwiki-user-profile-header">
+      <div class="media">
+        <div class="media-left">
+          <a href="$xwiki.getURL($reference)" title="$xwiki.getUserName($reference, false)">
```
mflorea, line 35: The title attribute value is missing XML escaping.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreference/view.vm
```diff
@@ -0,0 +1,78 @@
...
+        </div>
+        <div class="media-body">
+          <h2 class="media-heading">$xwiki.getUserName($reference)</h2>
+          <div class="xwiki-user-profile-comment">
+            #stripHTML($userPage.display('comment', 'view'))
+          </div>
+        </div>
+      </div>
+    </div>
+    #if ($!displayer.parameters.properties != '')
```
mflorea, line 47: AFAIK the silent notation is useful only when you are inside a string, such as ``"$!displayer.parameters.properties"``.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreference/view.vm
```diff
@@ -0,0 +1,78 @@
...
+              </li>
+            #end
+          #end
+        </ul>
+      </div>
+    #end
+  </div>
+#else
+  <div class="box warningmessage">
+    $services.localization.render('rendering.macro.userProfile.parameter.reference.isInvalid')
```
mflorea, line 76: Missing XML escape.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Translations.xml
```diff
@@ -29,6 +29,7 @@
   <creator>xwiki:XWiki.Admin</creator>
   <parent>xwiki:XWiki.Macros.WebHome</parent>
   <author>xwiki:XWiki.Admin</author>
+  <originalMetadataAuthor>XWiki.superadmin</originalMetadataAuthor>
```
mflorea, line 32: I think you need to omit this line as it's not supported / known by older versions of XWiki.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Translations.xml
```diff
@@ -105,7 +106,17 @@ rendering.macro.microsoftStream.parameter.start.description=Start video at a spe
...
+rendering.macro.microsoftStream.parameter.autoplay.description=Whether the video should start automatically or if the user must play it.
+
+## User Profile macro
+rendering.macro.userProfile.name=User Profile
+rendering.macro.userProfile.description=Displays a user profile with custom properties
+rendering.macro.userProfile.parameter.reference.name=User
+rendering.macro.userProfile.parameter.reference.description=Reference to a user page
+rendering.macro.userProfile.parameter.properties.name=User properties
+rendering.macro.userProfile.parameter.properties.description=List of user properties to be displayed, separated by a comma
+rendering.macro.userProfile.parameter.reference.isInvalid=The User Profile macro &lt;em&gt;reference&lt;/em&gt; parameter is either missing or invalid.
```
mflorea, line 118: You shouldn't use HTML (or wiki syntax) in a translation message. People helping with translation don't now HTML most of the time and the risk or breaking it (or even worse, inserting malicious HTML) is high. If the emphasize is really important then you should use:

```suggestion
rendering.macro.userProfile.parameter.reference.isInvalid=The User Profile macro {0}reference{1} parameter is either missing or invalid.
```

and pass the HTML wrapping a translation message parameters.

mflorea (changes requested):

slauriere:
> > I updated UserProfileMacroParameters to declare a default set of properties however I noticed that this makes it impossible to reset this property set to an empty list from the interface, could not figure out why yet.
> 
> The suggest input (based on Selectize) has code to cover precisely this use case, see https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-14.7/xwiki-platform-core/xwiki-platform-web/xwiki-platform-web-templates/src/main/resources/templates/macros.vm#L2732-L2736 . There's a hidden input with empty value added after the select, that is used to submit the empty selection when the select element has no option selected.

Right but the "properties" parameter is made editable via a simple input at the moment, not a Selectize. I'm wondering how I can control the behaviour of such a simple text input with an empty value.

slauriere:
> > It is styled by the stylesheet css/userprofile.css. That'd be better to use a LESS stylesheet so as to use some of the current theme colors but I could not find how to do this in this context for now. I will try to investigate, any hint welcome.
> 
> That's probably because currently `ssrx` doesn't support loading LESS files (i.e. doesn't support LESS compilation). I know that we support loading LESS files from WebJars (and compiling them in the context of the current skin), so one option is to add a WebJar module to the Pro Macros to hold client-side resources, but the issue is that this feature is relatively recent https://jira.xwiki.org/browse/XWIKI-18492 (13.3).

Thanks for the info. I'll add this as an improvement issue.

xwiki-pro-macros-api/pom.xml
```diff
@@ -49,5 +49,16 @@
       <groupId>org.webjars</groupId>
       <artifactId>jquery</artifactId>
     </dependency>
+    <!-- Dependencies needed by the User List macro -->
+    <dependency>
+      <groupId>org.xwiki.platform</groupId>
+      <artifactId>xwiki-platform-template-api</artifactId>
+      <version>${platform.version}</version>
+    </dependency>
```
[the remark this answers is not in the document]
slauriere: Indeed, it was not, I updated the POM with the dependencies actually used by the macro.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userprofile/internal/macro/UserProfileMacro.java
```diff
@@ -0,0 +1,94 @@
...
+    public List<Block> execute(UserProfileMacroParameters parameters, String content,
+        MacroTransformationContext context) throws MacroExecutionException
+    {
+
+        Map<String, String> params = new HashMap<>();
+        try {
+            String userReference = referenceSerializer.serialize(parameters.getReference());
+            params.put("reference", userReference);
+            params.put("properties", StringUtils.join(parameters.getProperties(), ','));
+            String html = htmlDisplayerManager.display(UserReference.class, content, params, "view");
```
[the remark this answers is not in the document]
slauriere: Absolutely. The updated code passes `parameters.getReference()` instead of the empty macro content.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreference/view.vm
```diff
@@ -0,0 +1,78 @@
...
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+## Hack copied from https://github.com/xwiki-contrib/application-changerequest/blob/application-changerequest-1.0.1/application-changerequest-ui/src/main/resources/ChangeRequest/Code/SaveChangeRequestModal.xml#L42
+#macro (stripHTML $displayOutput)
+  $stringtool.removeEnd($stringtool.removeStart($displayOutput, '{{html clean="false" wiki="false"}}'), '{{/html}}')
+#end
+#set ($ref = $displayer.parameters.reference)
+#set ($reference = $services.model.resolveDocument($ref))
```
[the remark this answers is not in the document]
slauriere: Updated accordingly as well.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreference/view.vm
slauriere: Right, now escaped.

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userreference/view.vm
slauriere: Thanks, I added the missing double quotes.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Translations.xml
slauriere: I removed this line indeed.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Translations.xml
slauriere: I understand, thanks for the explanation. I have replaced the emphasis by double quotes.

mflorea:
> Right but the "properties" parameter is made editable via a simple input at the moment, not a Selectize. I'm wondering how I can control the behaviour of such a simple text input with an empty value.

If you're using a text input (``<input type="text"/>``) then it's value (including the empty value) should be always submitted as long as the input is not disabled. Now, the Edit Macro dialog is a bit special, as there's no form there, we collect the values ourselves and we also check if the value is not the default value, in which case we omit it (i.e. we remove that parameter from the macro call).

mflorea (approved):

Issue #82, opened 2022-09-09
Create the "Attachments" macro


snazare:
Please create the "Attachments" macro, compatible with the Confluence Attachments macro: https://support.atlassian.com/confluence-cloud/docs/insert-the-attachments-macro/

Parameter | Default | Description
-- | -- | --
Filename Patterns(patterns) | all | A comma-separated list of regular expressions, used to filter the attachments by file name. Note that the parameter values must be regular expressions. For example:To match a file suffix of 'jpg', use .*jpg (not *.jpg).To match file names ending in 'jpg' or 'png', use .*jpg,.*png
Attachment Labels(labels) | (none) | A list of labels, used to filter the attachments to display. If you wish to enter more than one label, separate the labels with commas. Confluence will show only attachments that have all the labels specified. (The match is an AND, not an OR.) For information on labeling the attachments, see Use labels to organize pages and attachments.
Include Old Attachment Versions(old) | false | A value of true will include previous attachment versions in the list.
Sort By(sortBy) | date | The sort order for attachments. Note that people viewing the page can change the sort order by clicking the column headings. Valid values are:date – sorts by updated date in reverse chronological order (newest first)size – sorts largest to smallestname – sorts alphabeticallycreated date - sorts by creation date in reverse chronological order (newest first)
Sort Order(sortOrder) | ascending | Used in combination with the Sort By parameter, to sort the attachments in ascending or descending order.
Allow Upload(upload) | true | If selected, the list of attachments will include options allowing users to browse for, and attach, new files.
Page Title(page) | (none) | Used to display attachments from another page. If you do not enter a page title, the macro will display the files attached to the current page.
Show Previews(preview) | true | Used to display a preview of the attached file. If true, preview will be visible when the list item is expanded.

PR #87, opened 2022-10-03
Add Attachments macro (#82)

ClemDee:
This pull request adds support for the Attachment macro from confluence  (#82).

Possible improvements (discussed with @snazare, more input welcomed):
- Refresh attachment list when adding new attachments. There are two possibilities:
  - reload page automatically when adding new attachments (and remove the add button in edit mode)
  - use the dynamic attachments display from javascript
- Remove duplicate `addAttachments` html tag ids in page (due to reusing code from `attachments_macro.vm`)
- Add a parameter to display attachments from pages matching the given labels. Would work like the `page` parameter, but for several pages, and could be an alternative for the `labels` parameter which is not supported in XWiki.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
@@ -1,0 +1,944 @@
+<?xml version="1.1" encoding="UTF-8"?>
+
+<!--
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+-->
+
+<xwikidoc version="1.5" reference="Confluence.Macros.Attachments" locale="">
+  <web>Confluence.Macros</web>
+  <name>Attachments</name>
+  <language/>
+  <defaultLanguage/>
+  <translation>0</translation>
+  <creator>xwiki:XWiki.Admin</creator>
+  <parent>WebHome</parent>
+  <author>xwiki:XWiki.Admin</author>
+  <originalMetadataAuthor>XWiki.Admin</originalMetadataAuthor>
+  <contentAuthor>xwiki:XWiki.Admin</contentAuthor>
+  <version>1.1</version>
+  <title>Attachments</title>
+  <comment/>
+  <minorEdit>false</minorEdit>
+  <syntaxId>xwiki/2.1</syntaxId>
+  <hidden>true</hidden>
+  <content>The Attachment macro is a bridge between Confluence and XWiki. It uses the XWiki implementation to display attachments in the page content. There are two things to note:
+
+Not every parameters supported by the Confluence macro are supported by this bridge macro. The remaining parameters are ignored:
+
+* {{{ labels }}} is not supported since XWiki does not support attachment tags
+* {{{ preview }}} does not make sense for the XWiki macro as attachments are displayed differently
+* {{{ old }}} has not been implemented in this bridge macro because it does not make really sense for XWiki
+* {{{ "created date" }}} value of the {{{ sortBy }}} property behave the same way as the {{{ "date" }}} value
+
+= Parameters =
+
+|=Parameter|=Description|=Required|=Default
+|**patterns**|Comma-separated list of regular expressions, used to filter attachments by name.|No| 
+|**sortBy**|Sort attachments by {{{ date }}}, {{{ size }}} or {{{ name }}}|No|{{{ date }}}
+|**sortOrder**|Sort attachments in {{{ ascending }}} or {{{ descending }}} order|No|{{{ ascending }}}
+|**upload**|Allow users to attach new files|No|{{{ true }}}
+|**page**|Pages containing the attachments to display. Current page if empty.|No| 
+
+= Example Usage =
+
+
+{{code}}
+{{attachments
+  patterns=".*png"
+  sortBy="name"
+/}}
+{{/code}}</content>
+  <object>
+    <name>Confluence.Macros.Attachments</name>
+    <number>0</number>
+    <className>XWiki.StyleSheetExtension</className>
+    <guid>8a8f1c65-649d-41b2-ab48-778c676f24cd</guid>
+    <class>
+      <name>XWiki.StyleSheetExtension</name>
+      <customClass/>
+      <customMapping/>
+      <defaultViewSheet/>
+      <defaultEditSheet/>
+      <defaultWeb/>
+      <nameField/>
+      <validationScript/>
+      <cache>
+        <cache>0</cache>
+        <defaultValue>long</defaultValue>
+        <disabled>0</disabled>
+        <displayType>select</displayType>
+        <freeText>forbidden</freeText>
+        <largeStorage>0</largeStorage>
+        <multiSelect>0</multiSelect>
+        <name>cache</name>
+        <number>5</number>
+        <prettyName>Caching policy</prettyName>
+        <relationalStorage>0</relationalStorage>
+        <separator> </separator>
+        <separators>|, </separators>
+        <size>1</size>
+        <unmodifiable>0</unmodifiable>
+        <values>long|short|default|forbid</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </cache>
+      <code>
+        <contenttype>PureText</contenttype>
+        <disabled>0</disabled>
+        <editor>PureText</editor>
+        <name>code</name>
+        <number>2</number>
+        <prettyName>Code</prettyName>
+        <rows>20</rows>
+        <size>50</size>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.TextAreaClass</classType>
+      </code>
+      <contentType>
+        <cache>0</cache>
+        <disabled>0</disabled>
+        <displayType>select</displayType>
+        <freeText>forbidden</freeText>
+        <largeStorage>0</largeStorage>
+        <multiSelect>0</multiSelect>
+        <name>contentType</name>
+        <number>6</number>
+        <prettyName>Content Type</prettyName>
+        <relationalStorage>0</relationalStorage>
+        <separator> </separator>
+        <separators>|, </separators>
+        <size>1</size>
+        <unmodifiable>0</unmodifiable>
+        <values>CSS|LESS</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </contentType>
+      <name>
+        <disabled>0</disabled>
+        <name>name</name>
+        <number>1</number>
+        <prettyName>Name</prettyName>
+        <size>30</size>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </name>
+      <parse>
+        <disabled>0</disabled>
+        <displayFormType>select</displayFormType>
+        <displayType>yesno</displayType>
+        <name>parse</name>
+        <number>4</number>
+        <prettyName>Parse content</prettyName>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.BooleanClass</classType>
+      </parse>
+      <use>
+        <cache>0</cache>
+        <disabled>0</disabled>
+        <displayType>select</displayType>
+        <freeText>forbidden</freeText>
+        <largeStorage>0</largeStorage>
+        <multiSelect>0</multiSelect>
+        <name>use</name>
+        <number>3</number>
+        <prettyName>Use this extension</prettyName>
+        <relationalStorage>0</relationalStorage>
+        <separator> </separator>
+        <separators>|, </separators>
+        <size>1</size>
+        <unmodifiable>0</unmodifiable>
+        <values>currentPage|onDemand|always</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </use>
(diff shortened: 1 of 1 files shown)


ClemDee:
Changes from previous message:
- `addAttachment` button is no longer shown in page "edit" mode
- page is automatically refreshed when adding a new attachment
- added xwiki-only `tags` parameter used to display attachments from pages matching given tags. (for now this is union between tags, but intersection might make more sense, waiting input about it)

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+      <contentType>No content</contentType>
+    </property>
+    <property>
+      <defaultCategory>confluence</defaultCategory>
+    </property>
+    <property>
+      <description/>
+    </property>
+    <property>
+      <id>attachments</id>
```
mflorea, line 651: I'm not sure about this macro id. ATM we don't have in XWiki Standard an ``attachments`` macro but I'm pretty sure we'll have one in the near future, see https://jira.xwiki.org/browse/XWIKI-14631 , and then we'll have a conflict. I thought we were using a prefix for bridge macros in this repository, in order to prevent conflicts. This needs to be discussed with @snazare.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+      </type>
+    </class>
+    <property>
+      <defaultValue/>
+    </property>
+    <property>
+      <description>A comma-separated list of regular expressions, used to filter the attachments by file name. Note that the parameter values must be regular expressions. For example:
+
+* To match a file suffix of 'jpg', use .*jpg (not *.jpg).
+* To match file names ending in 'jpg' or 'png', use .*jpg,.*png
```
mflorea, line 735: Are you sure this looks good in the WYSIWYG editor Edit Macro modal? I'm asking because wiki syntax is not rendered there AFAIK. The macro parameter description is displayed as plain text. The line endings should also be lost because white spaces are collapsed in HTML (and ``\n`` doesn't break the line unless white spaces are preserved, but it's not the case here).

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+    <property>
+      <defaultValue>date</defaultValue>
+    </property>
+    <property>
+      <description>The sort order for attachments. Note that people viewing the page can change the sort order by clicking the column headings. Valid values are:
+
+* **date** – sorts by updated date in reverse chronological order (newest first)
+* **size** – sorts largest to smallest
+* **name** – sorts alphabetically
+* **created date** - sorts by creation date in reverse chronological order (newest first)</description>
```
mflorea, line 976: Same as above, I'm not sure this looks good in the Edit Macro modal from the WYSIWYG editor.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+        <size>60</size>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </type>
+    </class>
+    <property>
+      <defaultValue/>
+    </property>
+    <property>
+      <description>A list of comma-separated wiki tags, used to display attachments from pages matching all given tags. If you do not enter tags, the macro will display the files attached to the current page. When used, ~{~{~{ page }}} property is ignored.</description>
```
mflorea, line 1132: Again, wiki syntax in parameter description.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+#if ("$!wikimacro.parameters.tags" != '')
+  #set ($tags = $!wikimacro.parameters.tags.split(','))
+  #set ($pageReferenceSet = $collectiontool.getSet())
+  #foreach ($tag in $tags)
+    #set ($references = $xwiki.tag.getDocumentsWithTag("$tag"))
+    #set ($discard = $pageReferenceSet.addAll($references))
+  #end
+  #set ($attachments = [])
+  #foreach ($pageReference in $pageReferenceSet)
+    #set ($document = $xwiki.getDocument("$pageReference"))
```
mflorea, line 520:
```suggestion
    #set ($document = $xwiki.getDocument($pageReference))
```

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+      <async_context/>
+    </property>
+    <property>
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <code>{{html clean="false"}}
+  &lt;div class="confluence-attachment-macro"&gt;
+{{/html}}
+{{velocity output="false"}}
```
mflorea, line 505: I usually prefer to have the no-output block (that collects the data) at the start, in order to separate it from the block that displays the data.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+{{/warning}}
+#end
+#set ($attachmentsFiltered = $collectiontool.sort($attachmentsFiltered, "$sortBy:$sortOrder"))
+## Display attchments
+{{html clean="false"}}
+#if ($attachmentsFiltered.size() &gt; 0)
+  #displayAttachments($attachmentsFiltered)
+  #deleteAttachmentModal()
+#else
+  &lt;p class="noitems"&gt;$services.localization.render('core.viewers.attachments.noAttachments')&lt;/p&gt;
```
mflorea, line 591: Missing XML escaping for the translation value.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+#end
+{{/html}}
+## Allow upload
+#set ($upload = "$!wikimacro.parameters.upload")
+#if ($upload == 'true' &amp;&amp; ($hasEdit || $hasAdmin) &amp;&amp; $xcontext.action == 'view')
+
+
+  {{html}}
```
mflorea, line 599:
```suggestion
  {{html clean="false"}}
```
But I don't see why you start a new HTML macro when you can simply continue the previous one, no?

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+    &lt;form action="$doc.getURL("upload")" enctype="multipart/form-data" method="post" id="AddAttachment"&gt;
+    &lt;div&gt;
+    ## CSRF prevention
+    &lt;input type="hidden" name="form_token" value="$!{services.csrf.getToken()}" /&gt;
+    ## $redirect is defined in attachmentslist.vm
+    &lt;input type="hidden" name="xredirect" value="$escapetool.xml($redirect)" /&gt;
+    &lt;fieldset id="attachform"&gt;
+      &lt;legend&gt;$services.localization.render('core.viewers.attachments.upload.title')&lt;/legend&gt;
```
mflorea, line 607: Missing XML escaping.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+      ## Temporarily disabled, until we fix attachment name handling
+      ## &lt;div&gt;&lt;label id="xwikiuploadnamelabel" for="xwikiuploadname"&gt;$services.localization.render('core.viewers.attachments.upload.filename')&lt;/label&gt;&lt;/div&gt;
```
mflorea, line 609: Missing XML escaping. And line too long (would be good to keep it under 120 characters for improved readability).

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+      ## &lt;div&gt;&lt;input id="xwikiuploadname" type="hidden" name="filename" value="" size="40"/&gt;&lt;/div&gt;
+      &lt;div class="fileupload-field"&gt;
+        &lt;label class="hidden" for="xwikiuploadfile"&gt;$services.localization.render('core.viewers.attachments.upload.file')&lt;/label&gt;
```
mflorea, line 612: Missing XML escaping.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+        &lt;input id="xwikiuploadfile" type="file" name="filepath" size="40" class="uploadFileInput noitems" data-max-file-size="$!escapetool.xml($xwiki.getSpacePreference('upload_maxsize'))" /&gt;
+      &lt;/div&gt;
+      &lt;div&gt;
+        &lt;span class="buttonwrapper"&gt;
+          &lt;input type="submit" value="$services.localization.render('core.viewers.attachments.upload.submit')" class="button btn btn-primary"/&gt;
+        &lt;/span&gt;
+        &lt;span class="buttonwrapper"&gt;
+          &lt;a class="cancel secondary button btn btn-primary" href="$doc.getURL()"&gt;$services.localization.render('core.viewers.attachments.upload.cancel')&lt;/a&gt;
```
mflorea, line 620: Missing XML escaping.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+      &lt;/div&gt;
+    &lt;/fieldset&gt;
+    &lt;/div&gt;
+    &lt;/form&gt;
+  {{/html}}
+#end
+{{/velocity}}
+
+{{html clean="false"}}
+  &lt;/div&gt; &lt;!-- .confluence-attachment-macro --&gt;
```
mflorea, line 631: I'd rather keep this in the previous HTML macro and also inside Velocity, thus using a Velocity comment instead of generating a useless XML comment (on the client side).

mflorea (changes requested):

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -0,0 +1,1144 @@
...
+<xwikidoc version="1.5" reference="Confluence.Macros.Attachments" locale="">
+  <web>Confluence.Macros</web>
+  <name>Attachments</name>
+  <language/>
+  <defaultLanguage/>
+  <translation>0</translation>
+  <creator>xwiki:XWiki.Admin</creator>
+  <parent>WebHome</parent>
+  <author>xwiki:XWiki.Admin</author>
+  <originalMetadataAuthor>XWiki.Admin</originalMetadataAuthor>
```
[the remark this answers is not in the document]
tmortagne: The problem is just warning during install, but it's still not a good idea to include in pages stuff which are not supported by the minimum target XWiki version.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
tmortagne: I agree with @mflorea, I remember suggesting keeping the default "confluence_" prefix for macros coming from the confluence world to avoid any potential conflict and be very clear about what a macro is.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
ClemDee: Indeed, I figured this out too and I forgot to point it out!
So actually why are macro parameter description fields using the wysiwyg editor if they are rendered as plain text? From what you said I can expect this is not a bug but the intended behavior. But in that case this is quite confusing to have a wysiwyg editor here, especially considering that the macro description field is using a plain text editor on its side.

ClemDee: I'll try to update macro parameter descriptions to plain text, but this is going to be difficult for some of them which have rather complex/long descriptions like `sortBy`

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
ClemDee: Actually I took this from the `attachment_macros.vm`, so is there a reason why XML escaping is needed here and not in the vm, or is it missing too in the vm?

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
ClemDee: I create a PR on application-confluence-migrator in order to add the confluence attachment macro to the list of prefixed macros, see https://github.com/xwikisas/application-confluence-migrator/pull/79

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
mflorea: > why are macro parameter description fields using the wysiwyg editor

That's a bug IMO. See https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-store/src/main/java/org/xwiki/rendering/wikimacro/internal/WikiMacroClassDocumentInitializer.java#L77 versus https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-wikimacro/xwiki-platform-rendering-wikimacro-store/src/main/java/org/xwiki/rendering/wikimacro/internal/WikiMacroParameterClassDocumentInitializer.java#L55 . We should force the PURE_TEXT editor for parameter description field also.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
mflorea: We need to fix ``attachmentsinline.vm`` also. See https://jira.xwiki.org/browse/XWIKI-19749 .

mflorea (commented):
Two small comments left.

ClemDee:
Fixed two latest comments

mflorea (approved):
Thanks!

Issue #73, opened 2022-07-22
Create a Confluence bridge macro for Microsoft Stream Macro


oanalavinia:
Right now there is a Microsoft Stream Macro that gives the possibility to embed videos, see #69, but this is not compatible with the Confluence macro https://marketplace.atlassian.com/apps/1224419/microsoft-stream-for-confluence?tab=overview&hosting=cloud (see https://github.com/xwikisas/xwiki-pro-macros/issues/69#issuecomment-1192305122)

oanalavinia:
The macro should be added after https://jira.xwiki.org/browse/CONFLUENCE-105 is fixed, since right now the confluence migration will not copy the `url` macro parameter, which is mandatory.

Note that a bridge macro is needed in order to be compatible with the confluence macro name `net-presago-stream-macro`

oanalavinia:
https://jira.xwiki.org/browse/CONFLUENCE-105 has also been closed

PR #96, opened 2022-10-20
Create a Confluence bridge macro for Microsoft Stream Macro #73

oanalavinia:
* add the bridge macro with the confluence specific name and parameters
* small documentation update of the XWiki.Macros.MicrosoftStream macro page

The bridge macro only calls the already existing `msStream`. The differences between the two:
* the name, it being `net-presago-stream-macro` for the new one
* the name of the url parameter is `uri` for the bridge macro
* the alignment parameter is missing, since it's not present on confluence

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/MicrosoftStream.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/MicrosoftStream.xml
@@ -20,14 +20,14 @@
  * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
 -->
 
-<xwikidoc version="1.4" reference="XWiki.Macros.MicrosoftStream" locale="">
-  <web>XWiki.Macros</web>
+<xwikidoc version="1.4" reference="Confluence.Macros.MicrosoftStream" locale="">
+  <web>Confluence.Macros</web>
   <name>MicrosoftStream</name>
   <language/>
   <defaultLanguage/>
   <translation>0</translation>
   <creator>xwiki:XWiki.Admin</creator>
-  <parent>xwiki:XWiki.Macros.WebHome</parent>
+  <parent>WebHome</parent>
   <author>xwiki:XWiki.Admin</author>
   <contentAuthor>xwiki:XWiki.Admin</contentAuthor>
   <version>1.1</version>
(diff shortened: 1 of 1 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/MicrosoftStream.xml
```diff
@@ -0,0 +1,782 @@
...
+        <name>type</name>
+        <number>5</number>
+        <prettyName>Parameter type</prettyName>
+        <size>60</size>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </type>
+    </class>
+    <property>
+      <defaultValue/>
```
mflorea, line 689: Isn't there a default value? For boolean parameters it's usually easy to have a default value.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/MicrosoftStream.xml
```diff
@@ -0,0 +1,782 @@
...
```
mflorea, line 767: Same as above.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/MicrosoftStream.xml
```diff
@@ -0,0 +1,782 @@
...
+      <async_enabled>0</async_enabled>
+    </property>
+    <property>
+      <code>{{velocity}}
+#set ($uri = $escapetool.xml($wikimacro.parameters.get('uri')))
+#set ($width = $escapetool.xml($wikimacro.parameters.get('width')))
+#set ($height = $escapetool.xml($wikimacro.parameters.get('height')))
+#set ($showinfo = $escapetool.xml($wikimacro.parameters.get('showinfo')))
+#set ($start = $escapetool.xml($wikimacro.parameters.get('start')))
+#set ($autoplay = $escapetool.xml($wikimacro.parameters.get('autoplay')))
```
mflorea, line 282: We definitely need escaping in order to prevent code (wiki syntax, including script macros) injection, but why XML escaping? Is the ``msStream`` macro expecting XML escaped values for its parameters? That would be strange.

Now, indeed, we're missing https://jira.xwiki.org/browse/XWIKI-12839 so the current options are:
* perform the escaping yourself, such as https://github.com/xwiki/xwiki-platform/blob/xwiki-platform-14.8/xwiki-platform-core/xwiki-platform-panels/xwiki-platform-panels-ui/src/main/resources/Panels/Navigation.xml#L181-L182
* use ``$services.rendering.escape()`` but its behavior might change in the future because it isn't really designed for macro parameters, as Anca pointed out in https://jira.xwiki.org/browse/XWIKI-11979?focusedCommentId=88806&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-88806 .

@tmortagne do you have any recommendation on this?

mflorea (changes requested):

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/MicrosoftStream.xml
tmortagne: It's very unlikely that `$services.rendering.escape()` changes any time soon, as an optimized escaping would simply be much more complex (i.e. full of bugs) and much slower. As for "will is support wiki macro parameters in the future", since there are no indication of the context of what it's escaping it has to escape anything that can have a meaning in the indicated syntax and wiki macro syntax is wiki syntax.

mflorea: Then I think you should use ``$services.rendering.escape()``, @oanalavinia .

mflorea (approved):

Issue #85, opened 2022-09-11
Create the "Expand" bridge macro


snazare:
Please create the "Expand" bridge macro, compatible with the Confluence Expand macro: https://confluence.atlassian.com/doc/expand-macro-223222352.html

Please note that XWiki already has a similar macro: https://extensions.xwiki.org/xwiki/bin/view/Extension/Document%20Accordion%20Macro/

Parameters: 

Parameter | Default | Description
-- | -- | --
Title | Click here to expand... | Defines the text that appears next to the expand/collapse icon.

ClemDee:
Expand macro bridge implemented.

PR #104, opened 2022-12-23
Create the "Expand" bridge macro #85

ClemDee:


ClemDee:
you can test the macro on the Confluence dev server: https://confluence-migrator.devxwiki.com/xwiki/bin/view/DEMO/expand

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
```diff
@@ -0,0 +1,566 @@
...
+    <property>
+      <defaultCategory/>
+    </property>
+    <property>
+      <description>Add the Expand macro to your page to provide content in an expandable / collapsible section. 
+
+This is one of Confluence's most popular macros. It's great for:
+
+    visually reducing the amount of information on a page
+    breaking process information down into clickable steps
```
mflorea, line 466:
```suggestion
    visually reducing the amount of information on a page,
    breaking process information down into clickable steps,
```

Adding the comma ensures that the text is property read when the white space is stripped.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
```diff
@@ -0,0 +1,566 @@
...
+        <name>type</name>
+        <number>5</number>
+        <prettyName>Parameter type</prettyName>
+        <size>60</size>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </type>
+    </class>
+    <property>
+      <defaultValue>Click here to expand...</defaultValue>
```
mflorea, line 550: Should this text be translatable?

mflorea (changes requested):

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
```diff
@@ -0,0 +1,566 @@
...
+#set ($opened = $xcontext.action == 'edit')
+#set ($accorionId = "accordion-$expandMacroColllapseId")
+#set ($toggleId = "toggle-$expandMacroColllapseId")
+#set ($expandId = "collapse-$expandMacroColllapseId")
+
+{{html clean="false"}}
+&lt;div class="panel-group confluence-expand-macro" id="${accorionId}" role="tablist"&gt;
+  &lt;div class="panel panel-default"&gt;
+    &lt;div class="panel-heading" role="tab" id="${toggleId}"&gt;
+      &lt;h4 class="panel-title"&gt;
```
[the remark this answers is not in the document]
ClemDee: Not sure if it's absolutely needed, but as this is the title of the expand, I should be a good idea to indicate it semantically. The only issue I see with this is that the heading level is not going to be coherent with the rest of the page. In bootstrap, for that kind of titles (like modal titles), `h4` is used (and I saw this approach in other design systems too). So that's why I went for a `h4` for the title of this expand.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
ClemDee: How am I supposed to do that? I can't see a way to achieve this in the same manner as for macro name or description (as described [here](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/Tutorials/WritingMacros/WikiMacroTutorial/#HTranslations)) (I tried `rendering.macro.<macro id>.parameter.<parameter name>.defaultValue` but it didn't work).
I can't use `$services.localization.render` as the `{{wikimacroparameter name="title" /}}` will escape it.

What is the way to have a translated macro parameter default value?

mflorea: Indeed, currently there's no easy way to translate macro parameter default values. Let's keep it like this for now.

Issue #110, opened 2023-02-20
Add warning when adding macros without a license


oanalavinia:
Since macros are registered independently of an existing license, these can be added to a page even if the license does not yet exists. The macros are not fully functional in this case, because Licensing removes the view right on the macro page. 

A message should be displayed to users. This should be done at the code level of each macro.

oanalavinia:
Example https://github.com/xwikisas/application-diagram/blob/master/application-diagram-ui/src/main/resources/Diagram/DiagramMacro.xml#L327

PR #111, opened 2023-02-20
Add warning when adding macros without a license #110

oanalavinia:
* add license error for the macros defined in the _XWiki.Macros_ and _Confluence.Macros_ spaces
* move the code inside a `executeMacro` macro and call it only when there is a valid license
image

Note: for a better display of the differences, check the `hide whitespaces` in the `diff view` settings modal

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
@@ -507,7 +507,7 @@ Not every parameters supported by the Confluence macro are supported by this bri
     </property>
     <property>
       <code>{{velocity output="false"}}
-#macro (filterAndSortAttachments)
+#macro (filterAndSortAttachments $attachmentsFiltered $invalidSortBy $invalidSortOrder)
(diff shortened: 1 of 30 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -507,172 +507,190 @@ Not every parameters supported by the Confluence macro are supported by this bri
...
-  #set ($patterns = $!confluencePatterns.split(','))
-  #set($attachmentsFiltered = [])
-  #foreach ($attachment in $attachments)
-    #foreach ($pattern in $patterns)
-      #set ($matches = $attachment.filename.matches("(?i)$!pattern"))
-      #if ($matches)
-        #set ($discard = $attachmentsFiltered.add($attachment))
-        #break
-      #end
+#macro (filterAndSortAttachments)
```
mflorea, line 510: The macro doesn't declare any parameters but you call it below with parameters.

mflorea (approved):
Looks good with only one small comment.

Issue #113, opened 2023-02-21
The content of the columns from stacked user lists is not aligned


AndreeaChi:
Steps to reproduce:
1. Install Pro Macros 1.7
2. Create 2 users and complete the email, company, phone, address with content of different length
3. On a page, add a line of text and below it the user list macro with the properties: avatar,username,email,company,phone,address and one of the created users in the second field.
4. Add a second line of text and below it another user list macro with the same properties as above and the second created user.

Result:
<img width="563" alt="test-support-stacked-multiple-user-lists" src="https://user-images.githubusercontent.com/36840106/220363649-f3fae679-bf33-49fe-b28b-dd88877e8a74.png">

Wanted result would involve that the content is aligned even if the content in each cell differs.

AndreeaChi:
Once there are two users added in the user list, the content of the columns are automatically aligned within each list, providing a cleaner layout altogether:

<img width="539" alt="image" src="https://user-images.githubusercontent.com/36840106/220368030-e1c99945-7b4a-4128-a327-a40c79702f02.png">

raphj:
The CSS `table-layout: fixed` rule might help, though it comes with other issues. I believe one have to pick their poison in such situation, there's no ideal solution I think.

https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout

See also a potential problem related to fixed table layouts,
 with its solution https://stackoverflow.com/questions/17699324/fixed-table-layout-overflow-to-the-side

slauriere:
Thank you @raphj! Also adding here your suggestion for the record, I will consider it:

> I'm not totally sure this "fix" should be applied unconditionally, maybe a "fixed layout" macro parameter should be provided, or something like this.

PR #115, opened 2023-03-06
The content of the columns from stacked user lists is not aligned #113

slauriere:
- Introduce a new macro parameter "fixed table layout" (boolean, defaulting to false)
- Update userList CSS for ellipsing the text when a cell overflows

### DIFF
--- a/xwiki-pro-macros-api/src/main/resources/css/userlist.css
+++ b/xwiki-pro-macros-api/src/main/resources/css/userlist.css
@@ -1,3 +1,14 @@
-.xwiki-userlist-user-avatar {
-  width: 50px;
+.xwiki-userlist-fixed-layout {
+  table-layout: fixed;
+}
+
+/** Make sure table cells do not overflow when containing for instance long e-mail addresses. */
+.xwiki-userlist tr td {
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.xwiki-userlist tr td.xwiki-userlist-user-avatar {
+  width: 70px;
+  text-align: center;
 }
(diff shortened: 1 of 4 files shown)


AndreeaChi:
Hi @slauriere ! Could you please try with using a single user entry in each user list? The alignment problems usually were more visible in that situation.

slauriere:
Sure @AndreeaChi, here's another example with one user entry in each list. 

Here's the macro code for each of them (with different users), the first one has no address but a company filled in, while the second one has no company but an address.

```
{{userList properties="avatar,username,email,company,address,phone" users="XWiki.Alice" fixedTableLayout="true"/}}
```

users-lists

What do you think?

AndreeaChi:
Thank you for providing this example. It looks good!

xwiki-pro-macros-api/src/main/resources/css/userlist.css
```diff
@@ -1,3 +1,14 @@
-.xwiki-userlist-user-avatar {
-  width: 50px;
+.xwiki-userlist-fixed-layout {
+  table-layout: fixed;
+}
+
+/** Make sure table cells do not overflow when containing for instance long e-mail addresses. */
+.xwiki-userlist tr td {
```
mflorea, line 6: A ``td`` is always inside a ``tr`` so I'm not sure what's the benefit of specifying ``tr`` in the selector. Same below.

mflorea (approved):
Looks good, with one small comment.

xwiki-pro-macros-api/src/main/resources/css/userlist.css
slauriere: Right, I have removed the `tr` in both locations.

mflorea (approved):

Issue #112, opened 2023-02-21
Adapt the attachments macro to the new livetable attachments view


snazare:
In the recent versions of XWiki, we switched to livetable view for the attachments. This makes it that the current confluence_attachments macro is not functional anymore.

PR #123, opened 2023-04-14
Adapt the attachments macro to the new livetable attachments view #112

ClemDee:
* Adapt attachment macro to use the new livedata attachments view
* Remove `upload` parameter for now as it conflicts with standard attachments panel

There is also a slight issue with the `patterns` parameter. From the confluence side, this parameter expects a regex pattern in order to filter attachments. However, the livedata only supports `equals`, `startsWith` and `contains` filter operators for the filename field, so there is no direct equivalent to this regex filtering. This will make attachment macros to likely not display the expected files when that `patterns` parameter was set from imported confluence pages.

We might want to add  a new `pattern` operator to the attachment livedata source to address this issue.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
@@ -327,10 +327,6 @@ Not every parameters supported by the Confluence macro are supported by this bri
   #template('attachment_macros.vm')
   #template('display_macros.vm')
   #initRequiredSkinExtensions()
-  ## $xwiki.jsfx.use('uicomponents/widgets/upload.js', {'forceSkinAction': true, 'language': ${xcontext.locale}})
-  ## $xwiki.ssfx.use('uicomponents/widgets/upload.css', true)
-  ## $xwiki.jsfx.use('js/xwiki/viewers/attachments.js', {'forceSkinAction': true, 'language': ${xcontext.locale}})
-  ## $xwiki.jsx.use('Confluence.Macros.Attachments', {'language': ${xcontext.locale}})
   #set ($confluenceAttachmentMacroIndex = $confluenceAttachmentMacroIndex + 1)
   #set ($confluenceAttachmentMacroId = "confluenceAttachments$confluenceAttachmentMacroIndex")
   ##
@@ -409,7 +405,6 @@ Not every parameters supported by the Confluence macro are supported by this bri
   #end
 
   {{html clean="false" wiki="true"}}
-    ## #showConfluenceAttachmentsLiveData($document 'confluenceAttachments')
     #showConfluenceAttachments($document)
   {{/html}}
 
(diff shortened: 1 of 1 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -580,101 +311,108 @@ Not every parameters supported by the Confluence macro are supported by this bri
...
+#set ($confluenceAttachmentMacroIndex = -1)
+## Code taken and adapted
+#macro (showConfluenceAttachments $document)
+  #template('attachment_macros.vm')
+  #template('display_macros.vm')
+  #initRequiredSkinExtensions()
+  ## $xwiki.jsfx.use('uicomponents/widgets/upload.js', {'forceSkinAction': true, 'language': ${xcontext.locale}})
+  ## $xwiki.ssfx.use('uicomponents/widgets/upload.css', true)
+  ## $xwiki.jsfx.use('js/xwiki/viewers/attachments.js', {'forceSkinAction': true, 'language': ${xcontext.locale}})
+  ## $xwiki.jsx.use('Confluence.Macros.Attachments', {'language': ${xcontext.locale}})
```
mflorea, line 333: This looks like dead code so you should remove it.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Attachments.xml
```diff
@@ -580,101 +311,108 @@ Not every parameters supported by the Confluence macro are supported by this bri
...
-                  &lt;/a&gt;
-                &lt;/span&gt;
-              &lt;/div&gt;
-            &lt;/fieldset&gt;
-          &lt;/div&gt;
-        &lt;/form&gt;
-      #end
-    &lt;/div&gt; ## .confluence-attachment-macro
+  {{html clean="false" wiki="true"}}
+    ## #showConfluenceAttachmentsLiveData($document 'confluenceAttachments')
```
mflorea, line 412: Looks like dead code too.

mflorea (commented):
Looks good (minus the dead code).

ClemDee:
Dead code removed! :slightly_smiling_face:

ClemDee:
> Yes. For now you could use the pattern operator on your side (to be future-proof) and mention in the documentation that it doesn't work (i.e. it's a limitation of the XWiki core).

For what I can see, the Attachment Livedata is using using the Livetable source, so I guess it means that I would have to implement a custom source specially for the Attachment LD that implements this `pattern` operator, right?

mflorea (approved):

mflorea:
> For what I can see, the Attachment Livedata is using using the Livetable source, so I guess it means that I would have to implement a custom source specially for the Attachment LD that implements this pattern operator, right?

Yes, but I'm not sure it's worth the effort. What about converting the ``pattern`` constraint into ``startsWith`` or ``contains``. I guess it depends on the syntax used in the pattern. If it's just a wildcard, then it should be relatively easy to convert a pattern such as ``prefix*`` into ``startsWith:prefix``. Sure, you can't cover all cases, but maybe @snazare knows which patterns are the most used?

mflorea (approved):
We can improve the pattern filtering later, in a separate PR.

Issue #124, opened 2023-04-28
The Button macro deletes followup content when inserted from wysiwyg


ppantiru:
The html produced by wysiwyg seems to wrap around the followup content. 
![image](https://user-images.githubusercontent.com/8590004/235112569-4c0fa10a-757a-4470-82dc-35d83e3fc305.png)

lucaa:
This may be because of the button macro being invalid HTML / improperly nested, see https://jira.xwiki.org/browse/CKEDITOR-131 for more information about why the content always needs to be valid html.

PR #128, opened 2023-05-11
The Button macro deletes followup content when inserted from wysiwyg #124

ClemDee:
I used [the same fix than for the UserAvatar macro](https://github.com/xwiki/xwiki-platform/commit/be7047798f4029f570e137d3fc7a51fa974ba2f9), which is when the macro is not in inline mode, wrap its content inside a block to ensure inline content will be parsed correctly.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
@@ -323,7 +323,10 @@ $services.localization.render('rendering.macro.button.description')
   #darken($colors, $darkenedColors)
   {{html clean=false}}
   ## #124: The Button macro deletes followup content when inserted from wysiwyg
-  ## If not in inline mode, wrap inside block in order to fix issue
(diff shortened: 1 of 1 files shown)


xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
```diff
@@ -322,6 +322,9 @@ $services.localization.render('rendering.macro.button.description')
   #set ($darkenedColors = [])
   #darken($colors, $darkenedColors)
   {{html clean=false}}
+  ## #124: The Button macro deletes followup content when inserted from wysiwyg
+  ## If not in inline mode, wrap inside block in order to fix issue
```
mflorea, line 326:
```suggestion
  ## Wrap the macro output inside a block element if the context where the macro is inserted is not inline, e.g. if the
  ## macro is inserted between paragraphs rather than between the text of a paragraph. Otherwise the generated HTML is
  ## invalid which breaks the WYSIWYG editing because the browser tries to fix the invalid HTML and messes up the macro
  ## output markers that protect the macro output.
```

mflorea (approved):
The fix is good. I'd improve the comment.

Issue #144, opened 2023-06-09
Link not converted well when used within note macro


AndreeaChi:
After successfully importing a zip file on a 14.10.6 wiki with Confluence Migrator 1.0.14 and Pro Macros 1.9, on one of the resulting pages there is a note macro containing a link towards an external doc from a site present in the same Confluence space. However, it is rendered as a text.

See macro syntax in wiki mode:

<img width="676" alt="image" src="https://github.com/xwikisas/xwiki-pro-macros/assets/36840106/3530ea88-9b3f-458f-a396-8646dc27961b">

See rendered result:
<img width="613" alt="image" src="https://github.com/xwikisas/xwiki-pro-macros/assets/36840106/f12f9400-6464-4f2a-b421-82b5bb7ab213">

PR #145, opened 2023-06-12
Link not converted well when used within note macro #144

trrenty:
To fix this issue, I implemented the Note macro in java.


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/note/internal/macro/NoteMacro.java
```diff
@@ -0,0 +1,80 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xwiki.macros.note.internal.macro;
```
mflorea, line 20: Why the ".macro" suffix. Isn't it redundant if the package is prefixed with ``com.xwiki.macros``?

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/note/macro/NoteMacroParameters.java
```diff
@@ -0,0 +1,44 @@
...
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xwiki.macros.note.macro;
+
+/**
+ * @version $Id$
+ */
```
mflorea, line 24:
```suggestion
 * @since 1.10.1
 */
 @Unstable
```

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Note.xml
```diff
@@ -58,343 +58,4 @@ Result :
...
-#end
-{{/velocity}}
-
-{{include reference="Licenses.Code.VelocityMacros"/}}
-
-{{velocity}}
-## We need to check if there is a valid license because the macro is registered even if the user doesn't have view right
-## on the macro definition page. See XWIKI-14828: Rendering macros defined in wiki pages are available to users that
-## don't have view right on those pages.
-#if ($services.licensing.licensor.hasLicensureForEntity($xcontext.macro.doc.documentReference))
```
mflorea, line 282: This check seems to be lost in your implementation. This means you will be able to use the macro after the license expires.

mflorea (changes requested):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/note/internal/macro/NoteMacro.java
trrenty: I agree that it is redundant but two other macros already had this structure, `userlist` and `userprofile`. I went along with this. If I change it here, I should also change it on the other two. WDYT is the best option? to leave it as is or change them all. I think we should go with the latter.

mflorea: It's not a problem to rename the internal packages but for public packages this means an API breakage. Let's leave it like this for now, for consistency, it's not a big deal.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/AbstractProMacro.java
```diff
@@ -0,0 +1,133 @@
...
+        if (licensor.hasLicensure(
+            new DocumentReference(APP_WEBHOME, new WikiReference(wikiDescriptorManager.getCurrentWikiId()))))
+        {
+            return internalExecute(parameters, content, context);
+        } else {
+            String extensionName = localizationManager.getTranslationPlain("proMacros.extension.name");
+            String licenseAdminURL = accessBridge.getDocumentURL(
+                new DocumentReference(XWIKI_PREFERENCES, new WikiReference(wikiDescriptorManager.getMainWikiId())),
+                "admin", "editor=globaladmin&section=Licenses", "");
+            return Collections.singletonList(new MacroBlock(
```
mflorea, line 121: It's interesting that here you are returning directly an error macro block, while in ``NoteMacro`` you inject the warning macro and execute it. I'm not sure which option is best, but I'm wondering if you had any reason for doing it differently here and there? If there is a reason then maybe add a comment.

mflorea (commented):
Looks good. I only have a small comment / question :)

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/AbstractProMacro.java
trrenty: For the note macro, I initially returned a new macro block, as I am doing with the error message. As a result, when editing with the WYSIWYG editor, I would get the warning message macro injected in my wiki syntax. From this `{{note}}text{{/note}}` to this `{{note}}{{warning}}text{{/warning}}{{/note}}`. Every time i would edit the page, one more warning macro was added. This was because when instantiating a new warning macro, I passed the current content of the note macro. I'll add a comment describing this.

Issue #151, opened 2023-06-16
Recently Updated Macro displays an error on pages with different locales


trrenty:
Prerequisites:
In Global Administration > Content > Localization
1. Enable the wiki to be multilingual 
2. Add one more language to the supported languages
Steps to reproduce:
1. Create a page
3. Add a comment to the page
4. Add the Recently Updated Macro on the said page
5. View the page in the newly added supported language (by adding the query to the url `?language=de`
Result:
![image](https://github.com/xwikisas/xwiki-pro-macros/assets/45433221/7e6bc223-4e05-40f5-bb16-02e12d44d95e)

PR #152, opened 2023-06-16
Recently Updated Macro displays an error on pages with different loca…

trrenty:
…les #151

The issue was caused by the parsing of the comment date extracted from SOLR. Specifically, when parsing the date `Fri Jun 16 08:46:07 EEST 2023`, because the locale was `de` or something else, it was expecting `Freitag` instead of `Friday` to be present in the string.

I also ran the xar format.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdatedService.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdatedService.xml
@@ -216,8 +216,18 @@
   ## Add data to result
   #foreach ($property in $commentResponse.results)
     #if ($property.propertyname == 'date')
-      #set ($dateFormat = $datetool.getDateFormat("EEE MMM dd HH:mm:ss z yyyy", $datetool.getLocale(), $datetool.getTimeZone()))
-      #set ($date = $dateFormat.parse($property.propertyvalue_[0]))
+      ## We try to parse the date using the current locale. If if fails, we try with the 'en' locale. If that also
+      ## fails, the date won't be displayed.
+      #try('exception')
+        #set ($dateFormat = $datetool.getDateFormat("EEE MMM dd HH:mm:ss z yyyy", $datetool.getLocale(), $datetool.getTimeZone()))
+        #set ($date = $dateFormat.parse($property.propertyvalue_[0]))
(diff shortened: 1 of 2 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdatedService.xml
```diff
@@ -216,8 +216,18 @@
   ## Add data to result
   #foreach ($property in $commentResponse.results)
     #if ($property.propertyname == 'date')
-      #set ($dateFormat = $datetool.getDateFormat("EEE MMM dd HH:mm:ss z yyyy", $datetool.getLocale(), $datetool.getTimeZone()))
-      #set ($date = $dateFormat.parse($property.propertyvalue_[0]))
+      ## We try to parse the date using the current locale. If if fails, we try with the 'en' locale. If that also
+      ## fails, the date won't be displayed.
```
mflorea, line 220: It might be a good idea to set the date to null first, otherwise you may reuse a previous value.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdatedService.xml
```diff
@@ -216,8 +216,18 @@
...
+      #try('exception')
+        #set ($dateFormat = $datetool.getDateFormat("EEE MMM dd HH:mm:ss z yyyy", $datetool.getLocale(), $datetool.getTimeZone()))
```
mflorea, line 222: Too bad there was no comment explaining how this date format was chosen. In any case, you shouldn't duplicate it.

mflorea (changes requested):

mflorea:
The question you need to ask is why did you get a date with "Friday" when the locale is ``de``? Is this expected? If it is then sure, you ned to bulletproof your code, but if not then the problem may be elsewhere.

trrenty:
* I checked to see if the dates inside the result of the query were of type Date and unfortunately they were String. 
* I tested to see whether the format of the stored date of a comment is different depending on the locale of the document. It is the same format.
Taking these two points into consideration, I'll try to parse the date using only the 'en' locale and if it fails, it won't be displayed.

Issue #140, opened 2023-06-08
The avatars show with different size 


ane-gabriela:
Preconditions: Have some users in the wiki with different size images set as avatars.

Steps to reproduce:

1. Edit a page
2. Click on Insert > Other macros 
3. Search and select Team
4. Select "Show Usernames" and select a Size 
![TeamMacro](https://github.com/xwikisas/xwiki-pro-macros/assets/40691112/73dede7a-7f35-4826-9463-ad9566c2beca)
5. Click on Submit
6. Click on Save & View

Expected results:  The size of the avatars are the same in Edit and View mode.

Actual results:  The size (height and width) isn't the same for all avatars.
![TeamResults](https://github.com/xwikisas/xwiki-pro-macros/assets/40691112/97649175-37bb-457d-ab0e-95fb8bc3a739)

Environment: Windows 11, XWiki 14.10.10 with MySQL 8.0 and Chrome 114, Pro Macros 1.9

oanalavinia:
This was caused by a XWiki change https://jira.xwiki.org/browse/XWIKI-19432 , so was only on versions >= 13.10.4, 14.1

Farcasut:
Hi! This issue is happening again due to a regression caused by version 1.26.0.

PR #154, opened 2023-07-10
 The avatars show with different size #140 

Farcasut:
Fixed the issue where the avatar's height was determined by the size of the photo. The bug was caused by a change in 'https://github.com/xwiki/xwiki-platform/blame/master/xwiki-platform-core/xwiki-platform-flamingo/xwiki-platform-flamingo-skin/xwiki-platform-flamingo-skin-resources/src/main/resources/flamingo/less/general.less' where the img tag received the property height: auto;.  To fix this issue, I've overridden the height: auto; with the height given as a parameter to the macro.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
@@ -730,6 +730,7 @@
           title='$escapedDisplayUser'
           width='$escapedSize'
           height='$escapedSize'
+          style="height: ${escapedSize}px"
         /&gt;
       #end
       &lt;span class='xwikiteam-username'&gt;$escapedDisplayUser&lt;/span&gt;


oanalavinia:
@Farcasut the revert-layer is only compatible with Firefox https://developer.mozilla.org/en-US/docs/Web/CSS/revert-layer#browser_compatibility . Check the [XWiki Browser Support strategy](https://dev.xwiki.org/xwiki/bin/view/Community/SupportStrategy/BrowserSupportStrategy)

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -394,7 +394,8 @@
       <cache>long</cache>
     </property>
     <property>
-      <code>.xwikiteam .xwikiteam-ul {
+      <code>
```
oanalavinia, line 397: This is an unnecessary change.
Note that it's important that the changes in a commit to be only about that issue. This way the commit will be clean, easy to read and you won't make the `blame` feature harder to follow. So you need to check at the end the files changes, in case you forgot something in there.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -466,7 +467,6 @@
   align-self: center;
   padding: 1em;
 }
-
```
oanalavinia, line 469: Same as above

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -851,6 +851,14 @@
     #end
   #end
   {{html wiki=false clean=false}}
+
+  #set($escapedSize = $escapetool.xml($size))
```
oanalavinia, line 855: This needs better alignment.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -851,6 +851,14 @@
...
+        &lt;style&gt;
```
oanalavinia, line 856: It would be helpful to add a short comment explaining why this is needed. 
It makes sense in the context of this PR, but it might not be that obvious when someone else looks at the code

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -851,6 +851,14 @@
     #end
   #end
   {{html wiki=false clean=false}}
+
```
oanalavinia, line 854: Isn't this line influencing the macro display?

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -851,6 +851,15 @@
     #end
   #end
   {{html wiki=false clean=false}}
+  ##Starting with XWiki version 13.10.4 and 14.1-rc-1, img tags have a height: auto set and this breaks the macro fixed size feature. For this, a specific size is added
```
oanalavinia, line 854: You forgot a space after `##`
Same for below, it's `#set ($escap...`

Also, is this line longer than 120 characters? This is the characters limit used

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -851,6 +851,15 @@
...
+  ## https://jira.xwiki.org/browse/XWIKI-19432
```
oanalavinia, line 855: It is better to add the actual title of the jira issue. Like this, other devs don't need an extra click to know what the issue it's actual about and the comment is easier to read. E.g. https://github.com/xwikisas/application-licensing/blob/d168fb88fc0d121bf95e769ea21c55c00bebe5a6/application-licensing-licensor/application-licensing-licensor-api/src/main/java/com/xwiki/licensing/internal/upgrades/LicensingSchedulerListener.java#L51

oanalavinia (commented):
You still need to work a little on the alignment part. You could check https://dev.xwiki.org/xwiki/bin/view/Community/CodeStyle/ , but I know that it might not contain all the information

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
oanalavinia: Also, did you also tested the view on mobile? Asking because of https://jira.xwiki.org/browse/XWIKI-19432

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -852,6 +852,15 @@
   #end
   {{html wiki=false clean=false}}
   &lt;div class="xwikiteam #if(!$showUsernames)xwikiteam-usernames-hidden#end"&gt;
+  ## Starting with XWiki version 13.10.4 and 14.1-rc-1, img tags have a height: auto set and this breaks the macro fixed
+  ## size feature. For this, a specific size is added
+  ## Image not to scale on mobile XWIKI-19432
```
oanalavinia, line 857: A suggestion:
`Starting with XWiki 13.10.4 and 14.1, img tags have a 'height: auto' style set, see XWIKI-19432: Image not to scale on mobile. Since this breaks the macro fixed size feature, a specific size is added.`

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Team.xml
```diff
@@ -852,6 +852,15 @@
...
+  #set ($escapedSize = $escapetool.xml($size))
+  &lt;style&gt;
+    .fixedSize-$escapedSize {
+      height:${escapedSize}px;
```
oanalavinia, line 861: Missing space:
`height: ${escapedSize}px;`

oanalavinia (approved):

Issue #157, opened 2023-07-22
Viewpdf macro cannot be used inline


ClemensRobbenhaar:
The `viewpdf` macro cannot be used inline.
This seems obvious, as displaying a PDF inline makes no sense.

However the `viewpdf` can appear inline in pages migrated from confluence. The migrated pages then just show an error message.
A possible workaround is to allow this macro to be used "inline" by adding a container `(((...)))` around them. This of course interrupts the inline flow of the surrounding text, but makes he macro usable.  
As the likely reasons why the macro has been used inline from confluence is just a missing line break or the like, I think this is an acceptable solution.

PR is following.

ClemensRobbenhaar:
[See PR#158](https://github.com/xwikisas/xwiki-pro-macros/pull/158)

mflorea:
@ClemensRobbenhaar can you give us an example of XWiki syntax that is generated as a result of the Confluence migration where the ``viewpdf`` appears inline? I'd like to understand why the ``viewpdf`` is inline because, as you correctly noted, it doesn't make sense to have it inline.

The solution you propose is a workaround. You're avoiding the effect, but not fixing the root cause. Moreover, saying that the macro supports inline mode when it actually generates block level content (DIV, as a result of using the group wiki syntax ``(((...)))``) is not right. Same, the following IF is not semantically right:

    #if ($xcontext.macro.context.isInline())(((#end

This translates to "if the context is inline then generate a block level element", which doesn't make sense.

I'd like to understand why the ``viewpdf`` ends up inline after the migration, because this may be something that we can fix in the migrator / filter stream.

snazare:
Closing as inactive. We will need an example file to reproduce and fix the issue.

PR #158, opened 2023-07-22
issue157: Viewpdf macro cannot be used inline

ClemensRobbenhaar:
- add group markers if the macro is used inline to render the contained macros on block level

[See issue#157](https://github.com/xwikisas/xwiki-pro-macros/issues/157)


mflorea:
@raphj I'm not convinced by this fix. See my comment on https://github.com/xwikisas/xwiki-pro-macros/issues/157#issuecomment-1748343880 .

michitux:
Doesn't this kind of fix that lets an inline macro produce block-level content break the WYSIWYG editor as the resulting HTML is invalid and the browser's fixing it breaks the macro markers? See, e.g., https://jira.xwiki.org/browse/XRENDERING-517

Issue #138, opened 2023-06-07
File that has special chars in title can't be played


ane-gabriela:
Preconditions: Add a mp4 or mp3 file that has special chars in the title as attachment to the used page. Ex. 

https://github.com/xwikisas/xwiki-pro-macros/assets/40691112/75737976-6b97-42c8-9761-8cc797f549ed

Steps to reproduce:

1. Edit a page
2. Click on Insert > Other macros 
3. Search and select Multimedia Macro
4. Add the desired File name, the Page name (the file can be in a different page eg. 1.3.WebHome), width, height, autostart
5. Click on Submit
6. Click on Save & View
7. Refresh the page if the Video player takes long to load

Expected results: The mp4 / mp3 selected can be seen / heard. If autostart is selected the video starts when the user access the page, if not the play button must be clicked.

Actual results: The file is not working "Error loading media: File could not be played".
![Multimedia](https://github.com/xwikisas/xwiki-pro-macros/assets/40691112/dadb2656-49c8-4447-aacc-3d21ba57e2e5)

Environment: Windows 11, XWiki 14.10.10 with MySQL 8.0 and Chrome 114, Pro Macros 1.9

PR #162, opened 2023-09-13
File that has special chars in title can't be played #138

oanalavinia:
* escape special characters
* codestyle

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Multimedia.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Multimedia.xml
@@ -295,16 +295,20 @@ The result is the following :
     <property>
       <code>{{velocity output="false"}}
 #macro (executeMacro)
-  #set($attachment = $wikimacro.parameters.get('att--filename'))
-  #if(!$attachment)
-    #set($attachment = $wikimacro.parameters.get('name'))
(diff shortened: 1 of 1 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Multimedia.xml
```diff
@@ -295,16 +295,20 @@ The result is the following :
...
+    #set ($attachment = $services.rendering.escape(
+      $escapetool.css($wikimacro.parameters.get('name'), $xwiki.currentContentSyntaxId)))
   #end
-  #set($page = $wikimacro.parameters.get('page'))
-  #set($width = $wikimacro.parameters.get('width'))
-  #set($height = $wikimacro.parameters.get('height'))
-  #set($autostart = "$!wikimacro.parameters.get('autostart').toLowerCase()" == 'true')
-  #if("$!page" != '')
-    #set($attachment = "${page}@${attachment}")
+  #set ($page = $services.rendering.escape($wikimacro.parameters.get('page'), $xwiki.currentContentSyntaxId))
```
mflorea, line 304: See above. It doesn't make sense to escape this because you're not outputting directly its value. I'd rather escape the final value of ``attachment``.

mflorea (changes requested):

PR #168, opened 2023-10-12
[Misc] Improve macro parameter escaping - II

raphj:


xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
```diff
@@ -320,7 +320,7 @@ $services.localization.render('rendering.macro.button.description')
   #end
   #set ($darkenedColors = [])
   #darken($colors, $darkenedColors)
-  {{html clean=false}}
+  {{html clean=false wiki=false}}
```
michitux, line 323: `wiki=false` is the default, why are you setting this explicitly here?

oanalavinia (approved):

Issue #190, opened 2024-01-17
Excerpt-include macro does not display images from excerpt


ClemensRobbenhaar:
I.e if an excerpt contains an image:
![page_with_excerpt](https://github.com/xwikisas/xwiki-pro-macros/assets/4411046/8f0fe11b-46bc-44fc-987d-7d48211a08ae)

then this is displayed as:

![page_with_excerpt-include](https://github.com/xwikisas/xwiki-pro-macros/assets/4411046/988b1558-9280-4e43-a2a5-f3e8d95fbbe2)

as the current implementation uses "include"-style instead of "display"-style rendering.

I would also prefer if the macro is written in Java and registers its `0` parameter (the included page) as a reference, so renaming the referenced page should update that parameter.

acotiuga:
Hello @raphj! Please see the new PR https://github.com/xwikisas/xwiki-pro-macros/pull/210. 
The changes are 
`@since 1.14.4` -> `@since 1.14.5`
and 
```
TableBlock tableBlock = new TableBlock(List.of(
    new TableRowBlock(List.of(new TableHeadCellBlock(List.of(new WordBlock(document.getTitle()))))),
        new TableRowBlock(List.of(new TableCellBlock(Collections.singletonList(displayContent))))));

return List.of(tableBlock);
```
-> 
```
TableBlock tableBlock = new TableBlock(Arrays.asList(
    new TableRowBlock(
        Collections.singletonList(new TableHeadCellBlock(
            Collections.singletonList(new WordBlock(document.getTitle()))))),
    new TableRowBlock(
        Collections.singletonList(new TableCellBlock(
            Collections.singletonList(displayContent))))));

 return Collections.singletonList(tableBlock);
```
Second one ensure the successful build with Java 8.
Thanks for merging it and closing this issue and #192

PR #194, opened 2024-01-31
Excerpt-include macro does not display images from excerpt #190

acotiuga:
* migrate the macro from wiki page to java component
* fix the issue by displaying the content in it's own context.


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/macro/ExcerptIncludeMacro.java
```diff
@@ -0,0 +1,145 @@
...
+        DocumentReference reference = parameters.getReference();
+        XWikiContext xcontext = contextProvider.get();
+        XWikiDocument document;
+        try {
+            document = xcontext.getWiki().getDocument(reference, xcontext);
+        } catch (XWikiException e) {
+            throw new MacroExecutionException(String.format("Failed to get document for reference [%s]", reference), e);
+        }
+
+        if (!this.contextualAuthorization.hasAccess(Right.VIEW, reference)) {
```
oanalavinia, line 108: Should we also check for license? (e.g. https://github.com/xwikisas/application-confluence-migrator-pro/blob/a0ace06777aa4465144faa86e26f038c8fe75bce/application-confluence-migrator-pro-default/src/main/java/com/xwiki/confluencepro/internal/ConfluenceFilteringListener.java#L111 )

oanalavinia (commented):
I'm also not very familiar with this macro, but it looks ok to me. I only have one question

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/macro/ExcerptIncludeMacro.java
acotiuga: It is not necessary because the `internalExecute()` method is called in the `execute()` method from AbstractProMacro, which check for licensure https://github.com/xwikisas/xwiki-pro-macros/blob/main/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/AbstractProMacro.java#L115. 
I hope I'm not wrong.

oanalavinia: Indeed, didn't noticed it

Issue #200, opened 2024-02-08
Excerpt-Include macro reference is not properly resolved at migration


acotiuga:
Steps to reproduce:
1. Create in Confluence a space `Space` with 2 pages `Excerpt` (with excerpt macro) and `Excerpt-Include` (with excerpt-include macro referencing the Excerpt page)
2. Export from Confluence and do the migration in XWiki
3. The created pages are `Space.Excerpt.WebHome` and `Space.Excerpt-Include.WebHome`

Expected result:
1. The content of the `Space.Excerpt-Include.WebHome` should be `{{excerpt-include 0="Space.Excerpt.WebHome"/}}` similar to what is produced for the include macro `{{display reference="Space.Excerpt.WebHome"/}}`

Actual result:
1. The reference parameter, named `0` is not properly resolved and the content of the `Space.Excerpt-Include.WebHome` is `{{excerpt-include 0="Excerpt"/}}`.

acotiuga:
This is fixed now in Confluence Migrator Application (Pro) 1.11.0.

Issue #199, opened 2024-02-06
Excerpt-Include macro does not create backlinks on migration


acotiuga:
The problem is caused by the fact that the `excerpt-include` macro in Confluence uses the `defautl-parameter` parameter to store the reference which actually means there is no parameter name, and this is how it looks to have a parameter without name but with value: `{excerpt-include:My page name|nopanel=true}`.

Steps to reproduce:
1. Create a new page in Confluence with `excerpt` macro.
2. Create a new page in Confluence with `excerpt-include` macro, including the excerpt from the previous page.
3. Export the space and migrate it to XWiki

Actual result:
The XWiki page with the `excerpt` macro does not have a backlink to the XWiki page with the `excerpt-include` macro.
Expected result:
The XWiki page with the `excerpt` macro has a backlink to the XWiki page with the `excerpt-include` macro.

acotiuga:
Hi @michitux, @raphj! Here's a new PR for this issue: https://github.com/xwikisas/xwiki-pro-macros/pull/214. Thanks for reviewing it!

PR #203, opened 2024-02-09
Excerpt-Include macro does not create backlinks on migration #199

acotiuga:
Excerpt-Include macro reference is not properly resolved at import #200
* add a refactoring component to compute the correct reference in the excerpt-include 0 parameter -> A refactoring java class is needed to trigger the whole process done for Include and Display macros
* add a listener to watch when a page that has Excerpt macro is renamed to update the references in all the Excerpt-Include macro pages
* add a listener to watch when a page that has Excerpt-Include macro is renamed to update backlink on the included Excerpt page

### DIFF
--- a/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/ExcerptIncludeMacroRunnable.java
+++ b/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/ExcerptIncludeMacroRunnable.java
@@ -1,0 +1,136 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xwiki.macros.excerptinclude.internal;
+
+import java.util.List;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Provider;
+import javax.inject.Singleton;
+
+import org.slf4j.Logger;
+import org.xwiki.component.annotation.Component;
+import org.xwiki.model.reference.DocumentReference;
+import org.xwiki.model.reference.EntityReferenceSerializer;
+import org.xwiki.rendering.block.Block;
+import org.xwiki.rendering.block.XDOM;
+import org.xwiki.rendering.block.match.MacroBlockMatcher;
+import org.xwiki.stability.Unstable;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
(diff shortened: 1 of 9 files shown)


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/ExcerptIncludeMacroRunnable.java
```diff
@@ -0,0 +1,137 @@
...
+
+    @Inject
+    private Provider<XWikiContext> contextProvider;
+
+    @Inject
+    @Named("compact")
+    private EntityReferenceSerializer<String> compactEntityReferenceSerializer;
+
+    @Inject
+    private Logger logger;
```
michitux, line 63: This is hiding a field in the parent class and iirc in this case the component manager cannot inject the component. Did you test that this actually works?

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/listener/ExcerptIncludeMacroListener.java
```diff
@@ -0,0 +1,119 @@
...
+        XWikiDocument document = (XWikiDocument) source;
+        XWikiContext context = (XWikiContext) data;
+
+        List<Block> macroBlocks =
+            document.getXDOM().getBlocks(new MacroBlockMatcher("excerpt-include"), Block.Axes.DESCENDANT);
+
+        if (!macroBlocks.isEmpty()) {
+            try {
+                // We need to delete existing links before saving the page's ones.
+                storeHandler.deleteLinks(document.getId(), context);
```
michitux, line 96: Is it really this extension's job to delete all links for the document? Doesn't this also delete links that won't be re-added here? Also, why are links only deleted when there are macros - shouldn't they also be removed when there are no macros in case there was a macro before updating the document or when the document is deleted?

michitux (changes requested):
This implementation seems very fragile and error-prone. I would suggest to instead implement this by implementing a [MacroRefactoring](https://github.com/xwiki/xwiki-platform/blob/60b4f7f985560b36128630be9963f98464bf0df2/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/macro/MacroRefactoring.java) component, taking [IncludeMacroRefactoring](https://github.com/xwiki/xwiki-platform/blob/807db204d00a7591ddc9437de445e1c13edc538d/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-macros/xwiki-platform-rendering-macro-include/src/main/java/org/xwiki/rendering/internal/macro/include/IncludeMacroRefactoring.java#L60) as an example. This should lead to a much simpler implementation without an extra thread with a queue etc. - also documents won't be updated multiple times after a document move if there is both a link and an excerpt that need to be updated which could also lead to race conditions and errors. From all I could see, the necessary APIs should be available in XWiki 13.10.x.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/ExcerptIncludeMacroRunnable.java
ClemensRobbenhaar: I tested it, and having another logger here has the advantage that it logs for the current class and not the base class.
Then I have to admit that I did not check of the logger in the base class is usable.

michitux:
> The approach I used here is the same used previously in [Diagram Pro application](https://github.com/xwikisas/application-diagram/pull/140).

I think the use case of the diagram app is different: Diagrams aren't stored as macro, so using the mentioned macro refactoring is not possible. Further, the diagram is supposed to be the only content of the document afaik. Therefore, deleting all links and then adding the ones from the diagram seems appropriate. On the other hand, here the excerpt-include macro is just one of many macros in the page and thus deleting all links doesn't seem appropriate. Note also that in more recent versions of XWiki, the links table this code is modifying is not used anymore in most cases and instead the links that are stored in Solr are used instead.

In any case, I think this PR is a good starting point to write a `MacroRefactoring` component, so even if we should not use this as-is, thank you very much for the contribution!

PR #214, opened 2024-02-21
Excerpt-Include macro does not create backlinks on migration #199

acotiuga:


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/macro/ExcerptIncludeMacroRefactoring.java
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Implementation of reference refactoring operation for the excerpt-include macro.
+ *
+ * @version $Id$
+ * @since 1.14.6
+ */
+@Component
+@Singleton
+@Named("excerpt-include")
+@Unstable
```
michitux, line 53: I don't think there is any need to use `@Unstable` on an internal class.

michitux (approved):
Looks good to me, I guess you tested it and it works? I haven't verified that all this reference resolving is correct, but it looks plausible.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/excerptinclude/internal/macro/ExcerptIncludeMacroRefactoring.java
acotiuga: Indeed, I removed it.

Issue #213, opened 2024-02-21
The panel macro cuts off content that is too wide


michitux:
**Steps to reproduce:**

Create a panel macro that contains a very wide table. For example, take the following syntax:

```
{{panel}}
|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde|Content1Sedutperspiciatisunde
{{/panel}}
```

**Expected result:**

The table can be scrolled to see all columns.

**Actual result:**

The table cannot be scrolled, the content is just cut off. This is because the panel macro uses `overflow: hidden` on the border: https://github.com/xwikisas/xwiki-pro-macros/blob/1aac401b8154923d1e567a814d3ef246edabf6da/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Panel.xml#L215

PR #227, opened 2024-03-20
The panel macro cuts off content that is too wide #213

Farcasut:
Changed the the overflow property from hidden to scroll.
image


xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Panel.xml
```diff
@@ -212,7 +212,7 @@ This macro supports all Atlassian Confluence parameters as of version 7.9.</cont
 }
 .macro-border {
   margin-bottom: 10px;
-  overflow: hidden;
+  overflow: scroll;
```
michitux, line 215: From my understanding, this means that browsers will now always display a scrollbar, I don't think that this is desired. See also [MDN on overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#values).

trrenty (approved):

PR #236, opened 2024-04-02
General code improvement

Farcasut:
General code improvement

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Column.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Column.xml
@@ -435,9 +435,9 @@ Content in the column 4
   #set ($style = '')
   #set ($extraCssClass = 'withWidth')
   #if ("$!xcontext.macro.params.width" != '')
-    #set ($style = "style='width: $xcontext.macro.params.width'")
+    #set ($style = "style='width: $escapetool.xml($xcontext.macro.params.width)'")
   #end
-  (% class="macro-column" data-width="$xcontext.macro.params.width" $style %)(((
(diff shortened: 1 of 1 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Column.xml
```diff
@@ -435,9 +435,9 @@ Content in the column 4
   #set ($style = '')
   #set ($extraCssClass = 'withWidth')
   #if ("$!xcontext.macro.params.width" != '')
-    #set ($style = "style='width: $xcontext.macro.params.width'")
+    #set ($style = "style='width: $escapetool.xml($xcontext.macro.params.width)'")
   #end
-  (% class="macro-column" data-width="$xcontext.macro.params.width" $style %)(((
+  (% class="macro-column" data-width="$escapetool.xml($xcontext.macro.params.width)" $style %)(((
```
mflorea, line 440: ``$escapetool.xml()`` is not good here because you're not generating HTML (nor XML). You're generating wiki syntax. You need to use instead https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-rendering/xwiki-platform-rendering-xwiki/src/main/java/org/xwiki/rendering/script/RenderingScriptService.java#L219 , but only for the value (otherwise you escape the block parameter syntax)

    style="$!services.rendering.escape(..., $xwiki.currentContentSyntaxId)"

michitux: We also have a (relatively new) [developer guide about security](https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/Security/) that explains how to escape content in different situations.

Issue #246, opened 2024-04-23
Expand macro content is not displayed in a PDF export


lamblinc:
Content of an expand macro should be visible in a PDF export.

snazare:
I reproduce the issue on xwiki 15.10.10 Chrome 

`https://confluencemigration.demo.xwiki.com/xwiki/bin/get/Sandbox/?outputSyntax=plain&sheet=XWiki.PDFExport.WebHome&data=jobStatus&jobId=export%2Fpdf%2F1720441145082-102
`

PR #285, opened 2024-06-18
Expand Macro Content is Not Displayed in a PDF Export #246

Farcasut:
* Added a check to see if the page is in HTML print view; if we are in that view, we open the macro.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
@@ -409,10 +409,9 @@ Hello 👀
       <code>{{velocity output="false"}}
 #macro (executeMacro)
   $xwiki.ssx.use('Confluence.Macros.Expand')
-  #set ($opened = $xcontext.action == 'edit')
-  #set ($export = $request.getParameter('format'))
+  #set ($opened = $xcontext.action == 'edit' || $request.getParameter('format') == "html-print")
   {{html clean="false" wiki="true"}}
-  &lt;details class="confluence-expand-macro panel panel-default" #if ($opened || $!export == "html-print")open#end&gt;
+  &lt;details class="confluence-expand-macro panel panel-default" #if ($opened)open#end&gt;
     &lt;summary&gt;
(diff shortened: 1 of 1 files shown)


Farcasut:
Preview: 
image

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
```diff
@@ -410,9 +410,9 @@ Hello 👀
 #macro (executeMacro)
   $xwiki.ssx.use('Confluence.Macros.Expand')
   #set ($opened = $xcontext.action == 'edit')
```
trrenty, line 412:
```suggestion
  #set ($opened = $xcontext.action == 'edit' || $request.getParameter('format') == "html-print")
```

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
```diff
@@ -410,9 +410,9 @@ Hello 👀
...
-
+  #set ($export = $request.getParameter('format'))
   {{html clean="false" wiki="true"}}
-  &lt;details class="confluence-expand-macro panel panel-default" #if ($opened)open#end&gt;
+  &lt;details class="confluence-expand-macro panel panel-default" #if ($opened || $!export == "html-print")open#end&gt;
```
trrenty, line 415:
```suggestion
  &lt;details class="confluence-expand-macro panel panel-default" #if ($opened)open#end&gt;
```

trrenty (commented):
Also. Did you check both the "ctrl + p" function of the browser (that prints the page) and the pdf export functionality of XWiki? I dont think the first one is important for us. We are more interested in the pdf export.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
Farcasut: The Ctrl-P doesn't work, but the PDF export from XWiki works as intended now

Farcasut:
The Ctrl-P doesn't work, but the PDF export from XWiki works as intended now

Issue #295, opened 2024-07-09
The expand macro generates invalid HTML


oanat:
To be precise, it generates a DIV inside the summary element:
```
<details>
  <summary>
    <span class="panel-title">Title</span>
  </summary>
  <div class="panel-body">Body</div>
</details>
```
According to Marius the specification of the summary element says https://developer.mozilla.org/en-US/docs/Web/HTML/Element/summary

> The <summary> element's contents can be any heading content, plain text, or HTML that can be used within a paragraph.
> ...
> Permitted content: Phrasing content or one element of Heading content
> 

So DIV inside the summary element is not allowed. Replacing the DIV with a SPAN should fix the problem.

lucaa:
@Farcasut @trrenty shouldn't this ticket also have a milestone, now that it's closed? To know in what version to expect it fixed?

PR #297, opened 2024-07-09
The expand macro generates invalid HTML #295

Farcasut:
* Modified the content of the summary tag to be composed of a span instead of a div and added some css to preserve the old look.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
@@ -164,12 +164,11 @@ Hello 👀
   transform: rotate(90deg);
 }
 
-.confluence-expand-macro {
+.confluence-expand-macro .panel-title{
   padding: @panel-heading-padding;
+  display: block;
 }
-.panel-title {
-  padding: 5px;
-}
+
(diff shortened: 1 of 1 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
```diff
@@ -164,10 +164,12 @@ Hello 👀
...
 }
 
-.confluence-expand-macro .panel-title {
+.confluence-expand-macro {
   padding: @panel-heading-padding;
 }
-
+.panel-title {
+  padding: 5px;
+}
```
trrenty, line 172: Why is this needed?

The previous rule added a padding to the `.panel-title` inside a `.confluence-expand-macro`. Now there are two css rules. One applies some padding to `.confluence-expand-macro` and one applies padding to `.panel-title`.

Additionally, for the new rule, you use an absolute value for the padding, when, according to the previous rule, you can use a LESS variable.

trrenty: A `display: block` rule should suffice in order to have the same rendition.

trrenty (changes requested):

trrenty (approved):

Issue #206, opened 2024-02-13
Single-column layout doesn't cover page width


michitux:
**Steps to reproduce**

Create a page with the following content:

```
{{layout}}
{{layout-section ac:type="single"}}
{{layout-cell}}
{{info}}Some information{{/info}}
{{/layout-cell}}
{{/layout-section}}
{{/layout}}
```

**Expected result:**

The info box spans the whole page.

**Actual result:**

The info box is much smaller than the whole page.

![image](https://github.com/xwikisas/xwiki-pro-macros/assets/198317/e283a973-ef94-42fa-9aba-f8edfdc50fad)

Farcasut:
Hello, @michitux! Can you specify on which XWiki version you've encountered this issue and if it was a clean instance? I tried to replicate it on XWiki 13.10 and 14.10.1 and in both cases, I got the expected result.
13.10
![image](https://github.com/xwikisas/xwiki-pro-macros/assets/92780090/0e56ce21-ce2b-4b93-9b52-8e1ebc07c05a)
14.10.1
![image](https://github.com/xwikisas/xwiki-pro-macros/assets/92780090/5bea26e3-0357-4fba-9ee9-2e2575ef704c)

michitux:
I think this has been fixed in the meantime by 65aa0370606f0faed7eb432b724126d818ee533b, I cannot reproduce after upgrading the pro macros. However, I wonder if the whitespace left of the content is desired.

snazare:
Tested on 15.10.10 pro macros 1.18.0 - indeed the box takes most screen but there is an odd empty space at the beginning of the info box. @Farcasut would you have the chance to fix this?

PR #299, opened 2024-07-15
Single-column layout doesn't cover page width #206.

Farcasut:
The issue was caused by a CSS class that was meant to leave a gap between the columns, but it overlooked the fact that the rule would indent the first column. To resolve this, I've removed the margin-left CSS and added a gap to the flex element so the columns have the same space between them.

Preview: 
image


xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceLayoutSection.xml
```diff
@@ -192,12 +192,12 @@ This is a bridge for the Confluence Layout Section macro. It is usually used wit
...
 
 .macro-layout-section {
+  gap: 32px;
   h1, h2, h3, h4, h5, h6 {
     margin: 0;
   }
 }
 .macro-layout-cell {
-  margin-left: 32px;
 }</code>
```
trrenty, line 201: You can also remove the empty css rule

trrenty (approved):

PR #313, opened 2024-07-29
Conversion from Confluence diagram fails without visible error message

ClemensRobbenhaar:
- in case the XML for the diagram is empty, return an error
- handle case where the `diagram` element in the XML has no further attributes

Fixes [application-diagram#276](https://github.com/xwikisas/application-diagram/issues/276) (which I reported in the wrong project)


xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/DiagramService.xml
```diff
@@ -64,22 +64,29 @@
       #getUniquePageReference($xwiki.getDocument($page).space $uniquePageReference)
       #set($diagramContent = $diagramAttach.getContentAsString())
       #set($diagramPNGContent = $diagramPreviewAttach.getContentAsString())
-      #if($diagramContent.contains('&lt;mxfile ') &amp;&amp; $diagramContent.contains('&lt;diagram '))
+      #if($diagramContent.contains('&lt;mxfile ') &amp;&amp; ($diagramContent.contains('&lt;diagram ') || $diagramContent.contains('&lt;diagram&gt;')))
```
acotiuga, line 67: If the XML tag for the diagram is empty (no children), shouldn't there be a case to handle `<diagram/>` ?

Issue #342, opened 2024-09-05
Add a new parameter to the panel macro called height


Farcasut:
Improve the panel macro by adding a new optional parameter called height.
Related issue: https://github.com/xwikisas/application-confluence-migrator-pro/issues/171

PR #343, opened 2024-09-05
 Add a new parameter to the panel macro called height #342

Farcasut:
* New parameter called height


xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Panel.xml
```diff
@@ -221,6 +222,7 @@ This macro supports all Atlassian Confluence parameters as of version 7.9.</cont
 
 /*default border style*/
 .macro-panel.macro-border, .macro-panel &gt; .macro-border {
+  display:grid;
```
trrenty, line 225:
```suggestion
  display: grid;
```

PR #345, opened 2024-09-16
Expand extra param

Farcasut:


trrenty:
I don't think "visible" is a good name for the parameter. Since we are talking about the "expand" macro, it should either be "expanded", "opened", "collapsed" or something along the lines. "visible" makes me think that you want to hide/unhide the expand macro.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Translations.xml
```diff
@@ -89,8 +89,8 @@ rendering.macro.panel.parameter.classes.description= If specified adds css class
 
 ## Expand macro
 rendering.macro.expand.defaultexpandtitle=Click here to expand...
-rendering.macro.expand.parameter.visible.name=Visible
-rendering.macro.expand.parameter.visible.description=If selected the expand will be opened automatically when loading the page.
+rendering.macro.expand.parameter.expanded.name=Visible
```
trrenty, line 92: Expanded

Issue #352, opened 2024-10-23
Implement a macro for tab


Josue-T:
This is to implement an equivalent macro of [Deck of cards](https://appfire.atlassian.net/wiki/spaces/CTFCSM/pages/471176665/Deck+of+Cards) and [card](https://appfire.atlassian.net/wiki/spaces/CTFCSM/pages/471176639/Card) macro from confluence.

PR #353, opened 2024-10-23
#352: Implement a macro for tab

Josue-T:
Implement `tab` and `tab-group` macro 
cf. https://github.com/xwikisas/xwiki-pro-macros/issues/352

### DIFF
--- a/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
+++ b/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
@@ -1,0 +1,203 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xwiki.macros.tab.internal;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Random;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Provider;
+import javax.inject.Singleton;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.rendering.block.Block;
+import org.xwiki.rendering.block.BulletedListBlock;
+import org.xwiki.rendering.block.GroupBlock;
+import org.xwiki.rendering.block.ListItemBlock;
+import org.xwiki.rendering.block.RawBlock;
+import org.xwiki.rendering.block.match.MacroBlockMatcher;
+import org.xwiki.rendering.macro.MacroContentParser;
+import org.xwiki.rendering.macro.MacroExecutionException;
+import org.xwiki.rendering.macro.descriptor.DefaultContentDescriptor;
+import org.xwiki.rendering.syntax.Syntax;
+import org.xwiki.rendering.transformation.MacroTransformationContext;
+import org.xwiki.skinx.SkinExtension;
+import org.xwiki.text.StringUtils;
+import org.xwiki.xml.XMLUtils;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xwiki.macros.AbstractProMacro;
+import com.xwiki.macros.tab.macro.TabGroupMacroParameters;
+import com.xwiki.macros.tab.macro.TabMacroParameters;
+
+/**
+ * Tab-group macro: the main container macro for tab macro elements.
+ *
+ * @version $Id: $
+ * @since 1.24.0
+ */
+@Component
+@Named("tab-group")
+@Singleton
+public class TabGroupMacro extends AbstractProMacro<TabGroupMacroParameters>
+{
+    private static final String NAME = "Tab group";
+
+    private static final String DESCRIPTION = "The main macro which group tab macro elements.";
+
+    private static final String CONTENT_DESCRIPTION =
+        "It's expected that the content of this macro are only tab macro. "
+            + "Note all other element will be ignored. Only tab macro element will be rendered.";
+
+    private static final String TRUE = "true";
+
+    private static final String FALSE = "false";
+
+    private static final String ID = "id";
+
+    private static final String TAB_MACRO_ID = "tab";
+
+    private static final String TAB_MACRO_PARAM_LABEL = "label";
+
+    private static final String CSS_PX_UNIT = "px;";
+
+    private static final String BLOCK_PARAM_CLASS = "class";
+
+    private static final String BLOCK_PARAM_ROLE = "role";
+
+    private static final Random RANDOM = new Random();
+
+    @Inject
+    private Provider<XWikiContext> xwikiContextProvider;
+
+    @Inject
+    private MacroContentParser contentParser;
+
+    @Inject
+    @Named("ssrx")
+    private SkinExtension ssfx;
+
+    /**
+     * Create and initialize the descriptor of the macro.
+     */
+    public TabGroupMacro()
+    {
+        super(NAME, DESCRIPTION, new DefaultContentDescriptor(CONTENT_DESCRIPTION, true, Block.LIST_BLOCK_TYPE),
+            TabGroupMacroParameters.class);
+    }
+
+    @Override
+    public boolean supportsInlineMode()
+    {
+        return false;
+    }
+
+    private List<Block> renderView(TabGroupMacroParameters parameters, String content,
+        MacroTransformationContext context) throws MacroExecutionException
+    {
+        ssfx.use("css/tabmacro.css");
+        String macroId = parameters.getId();
+
+        if (StringUtils.isEmpty(macroId)) {
+            // FIXME we should have a better way to generate an ID
+            macroId = "tab-group-" + RANDOM.nextInt();
+        }
+
+        Block contentBlocks = this.contentParser.parse(content, context, false, context.isInline());
+
+        List<Block> ulBlocks = new ArrayList<>();
+        List<Block> macroBlocks = contentBlocks.getBlocks(new MacroBlockMatcher(TAB_MACRO_ID), Block.Axes.CHILD);
+        Optional<Block> defaultTabBlock = macroBlocks.stream()
+            .filter(i -> TRUE.equals(i.getParameter(TabMacroParameters.PARAM_NAME_DEFAULT)))
+            .findFirst();
+        int defaultTabIndex = defaultTabBlock.isPresent() ? macroBlocks.indexOf(defaultTabBlock.get()) : 0;
+
+        int inc = 0;
+        for (Block mb : macroBlocks) {
+            boolean isActive = inc == defaultTabIndex;
+            mb.setParameter(TabMacroParameters.PARAM_NAME_DEFAULT, isActive ? TRUE : FALSE);
+            String id = mb.getParameter(ID);
+            if (id == null) {
+                id = macroId + "_" + inc;
+                mb.setParameter(ID, id);
+            }
+            Map<String, String> lbParam = new HashMap<>();
+            lbParam.put(BLOCK_PARAM_ROLE, "presentation");
+            if (isActive) {
+                lbParam.put(BLOCK_PARAM_CLASS, "active");
+            }
+            /*
+            For now, I use the raw block for these reason:
+            - the LinkBlock generate a span element which break bootstrap tabs
+            - the word block can't contain a special char - but well this way it probably not better.
+             */
+            RawBlock linkBlock = new RawBlock(
+                String.format("<a href=\"#%s\" aria-controls=\"%s\" role=\"tab\" data-toggle=\"tab\">%s</a>",
+                    id, id, XMLUtils.escape(mb.getParameter(TAB_MACRO_PARAM_LABEL))),
+                Syntax.HTML_5_0);
+            ulBlocks.add(new ListItemBlock(Collections.singletonList(linkBlock), lbParam));
+            inc++;
+        }
+        Map<String, String> mainBlockParam = new HashMap<>();
+        StringBuilder style = new StringBuilder();
+        StringBuilder mainDivClass = new StringBuilder("xwikitabmacro");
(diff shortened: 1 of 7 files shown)


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
```diff
@@ -0,0 +1,203 @@
...
+            }
+            /*
+            For now, I use the raw block for these reason:
+            - the LinkBlock generate a span element which break bootstrap tabs
+            - the word block can't contain a special char - but well this way it probably not better.
+             */
+            RawBlock linkBlock = new RawBlock(
+                String.format("<a href=\"#%s\" aria-controls=\"%s\" role=\"tab\" data-toggle=\"tab\">%s</a>",
+                    id, id, XMLUtils.escape(mb.getParameter(TAB_MACRO_PARAM_LABEL))),
+                Syntax.HTML_5_0);
```
Josue-T, line 161: It's probably not the best way to do this. Do you have any idea how we can implement this in a cleaner way ?

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
```diff
@@ -0,0 +1,203 @@
...
+
+    private List<Block> renderView(TabGroupMacroParameters parameters, String content,
+        MacroTransformationContext context) throws MacroExecutionException
+    {
+        ssfx.use("css/tabmacro.css");
+        String macroId = parameters.getId();
+
+        if (StringUtils.isEmpty(macroId)) {
+            // FIXME we should have a better way to generate an ID
+            macroId = "tab-group-" + RANDOM.nextInt();
```
Josue-T, line 127: Does we have a good solution to ensure that we have a unique ID on a page ?

xwiki-pro-macros-api/src/main/resources/css/tabmacro.css
```diff
@@ -0,0 +1,3 @@
+body.content.preference-underlining-only-inline-links #xwikicontent div.xwikitabmacro ul.nav.nav-tabs li a {
```
trrenty, line 1: I think the this selector is overly specific. 

I think this should suffice `div.xwikitabmacro ul.nav.nav-tabs li a`. Its possible to have xwiki syntax in places other than the xwiki content.

trrenty (approved):
Left some comments. You can merge after going over them

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
Josue-T: Thanks how to ensure that we use the same instance of `IdGenerator` for all macro of the same page, but a new one for each page ?

trrenty: https://github.com/xwiki/xwiki-platform/blob/master/xwiki-platform-core/xwiki-platform-display/xwiki-platform-display-macro/src/main/java/org/xwiki/rendering/internal/macro/display/DisplayMacro.java#L113-L114

This piece of code suggests that the MacroTransformationContext should contain one. Im not sure how it works, but i reckon that the TransformationContext should represent the page that we are in. Otherwise, maybe we could get our hands on the XDOM of the current  page (where the macro is being executed) which should provide an IdGenerator.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
Josue-T: Following [the discussion that we had on matrix](https://matrix.to/#/!ikPtGZaGWtyblizzlR:matrix.xwiki.com/$17302157451120tahEz:matrix.xwiki.com?via=matrix.org&via=matrix.xwiki.com&via=mtrx.nz) it seem not really possible work around the span element. So the 2 solution that we have, are:
- keep the RawBlock element but it's not ideal as we lost some xwiki features.
- fix the issue, when there are a span around the link.

After a test, the issue is releated to the CSS part. The JS part still work if we use link block.

After some investiguation, it's probably possible to fix the theme to make it working with span but it might be difficult to support all flaming theme this way as in some theme there are customization for tab element.

On other side, the tab element without span is used at many place in XWiki, so to me the argument that we lost xwiki features if we don't wrap link in a span is not really important (if it was really an issue we should fix all tab in xwiki).
So my conclusion would be to keep it as it is with the rawBlock, which is not ideal but is the most simple solution in this context.

trrenty: Lets keep it as it is then. Thank u for the investigation

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
```diff
@@ -0,0 +1,203 @@
...
+            inc++;
+        }
+        Map<String, String> mainBlockParam = new HashMap<>();
+        StringBuilder style = new StringBuilder();
+        StringBuilder mainDivClass = new StringBuilder("xwikitabmacro");
+        if (!StringUtils.isEmpty(parameters.getId())) {
+            mainBlockParam.put(ID, parameters.getId());
+        }
+        if (parameters.getWidth() > 0) {
+            style.append("width: ").append(parameters.getWidth()).append(CSS_PX_UNIT);
```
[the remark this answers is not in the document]
Josue-T: Converted the parameter to string and so the user should put the unit in the parameter like `20px`. It was implemented this way in the other knowledge-management products.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
```diff
@@ -0,0 +1,253 @@
...
+
+            Map<String, String> lbParam = new HashMap<>();
+            lbParam.put(BLOCK_PARAM_ROLE, "presentation");
+            if (isActive) {
+                lbParam.put(BLOCK_PARAM_CLASS, "active");
+            }
+            // We use the raw block because the LinkBlock generate a span element which break bootstrap tabs CSS
+            RawBlock linkBlock = new RawBlock(
+                String.format("<a href=\"#%s\" aria-controls=\"%s\" role=\"tab\" data-toggle=\"tab\">%s</a>",
+                    id, id, XMLUtils.escape(mb.getParameter(TAB_MACRO_PARAM_LABEL))),
```
michitux, line 178: From what I can see, the id is a parameter and thus needs escaping.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
```diff
@@ -0,0 +1,253 @@
...
+        StringBuilder style = new StringBuilder();
+        StringBuilder mainDivClass = new StringBuilder("xwikitabmacro");
+        if (!StringUtils.isEmpty(parameters.getId())) {
+            mainBlockParam.put(ID, parameters.getId());
+        }
+        if (!StringUtils.isEmpty(parameters.getWidth())) {
+            style.append("width: ").append(parameters.getWidth());
+        }
+        if (!StringUtils.isEmpty(parameters.getHeight())) {
+            style.append("height: ").append(parameters.getWidth());
```
michitux, line 193: This looks like a typo, shouldn't this be

```suggestion
            style.append("height: ").append(parameters.getHeight());
```

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
```diff
@@ -0,0 +1,253 @@
...
+            ulBlocks.add(new ListItemBlock(Collections.singletonList(linkBlock), lbParam));
+            inc++;
+        }
+        Map<String, String> mainBlockParam = new HashMap<>();
+        StringBuilder style = new StringBuilder();
+        StringBuilder mainDivClass = new StringBuilder("xwikitabmacro");
+        if (!StringUtils.isEmpty(parameters.getId())) {
+            mainBlockParam.put(ID, parameters.getId());
+        }
+        if (!StringUtils.isEmpty(parameters.getWidth())) {
```
michitux, line 189: While not wrong, I think it would be nicer to use the method [`isNotEmpty`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isNotEmpty(java.lang.CharSequence))

```suggestion
        if (StringUtils.isNotEmpty(parameters.getWidth())) {
```

(same for all other parameters)

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabGroupMacro.java
```diff
@@ -0,0 +1,253 @@
...
+            RawBlock linkBlock = new RawBlock(
+                String.format("<a href=\"#%s\" aria-controls=\"%s\" role=\"tab\" data-toggle=\"tab\">%s</a>",
+                    id, id, XMLUtils.escape(mb.getParameter(TAB_MACRO_PARAM_LABEL))),
+                Syntax.HTML_5_0);
+            ulBlocks.add(new ListItemBlock(Collections.singletonList(linkBlock), lbParam));
+            inc++;
+        }
+        Map<String, String> mainBlockParam = new HashMap<>();
+        StringBuilder style = new StringBuilder();
+        StringBuilder mainDivClass = new StringBuilder("xwikitabmacro");
```
michitux, line 185: Wouldn't it be nicer to collect the classes in a `List` and to then join them at the end?

xwiki-pro-macros-api/src/main/resources/js/tabmacro.js
```diff
@@ -0,0 +1,56 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+const delay = (delayInms) => {
```
michitux, line 20: Why is this code defining a global method?

xwiki-pro-macros-api/src/main/resources/js/tabmacro.js
```diff
@@ -0,0 +1,56 @@
...
+    let tabMacroElements = document.querySelectorAll('div.xwikitabmacro');
+    try {
+      let promises = tabMacroElements.values().map(i => processTabGroupElement(i));
+      Promise.all(promises).catch((err) => console.log(err));
+    }
+    catch (err) {
+      console.error(err);
+    }
+  });
+});
```
michitux, line 56: This code could use some comments, it is super hard to understand, and it is not clear to me why it does things the way it does them. For example, why does it introduce a global method `delay`? Why is delay returning a promise that just calls `setTimeout`? Why is this async and not just a regular method? Also, it seems weird to me that the whole configuration is serialized as JSON if only two values out of the configuration are actually needed, it would seem much more natural to me to just store these two values as data attributes.

michitux (changes requested):

xwiki-pro-macros-api/src/main/resources/js/tabmacro.js
Josue-T: it's not needed. Should I move it into the `require(...)` block ?

xwiki-pro-macros-api/src/main/resources/js/tabmacro.js
Josue-T: > For example, why does it introduce a global method delay?

I'll inline it we don't really need it.

> Why is this async and not just a regular method ?

Because I feel that it was the easiest way to implement this. But maybe there are simpler way to do this.

> Also, it seems weird to me that the whole configuration is serialised as JSON if only two values out of the configuration are actually needed, it would seem much more natural to me to just store these two values as data attributes.

Because I thought to extends the feature of this macro it's easier to have all configuration in the js part. But yes I agree that now it's a bit overkill. I'll do what you suggested.

michitux (changes requested):
I found two more issues, apart from that it looks good.

xwiki-pro-macros-api/src/main/resources/js/tabmacro.js
```diff
@@ -0,0 +1,52 @@
...
+
+    let globalNextAfter = parseInt(element.getAttribute('data-nextafter'));
+    let loopCards = element.getAttribute('data-loopcards');
+    let elementToIterate = element.querySelectorAll("ul.nav-tabs li a");
+
+    if (!isNaN(tabNextAfter) && globalNextAfter > 0 && elementToIterate.length > 0) {
+      for (let i = 0; i < 1 || loopCards === "true"; i++) {
+        for (const el of elementToIterate) {
+          let divElement = element.querySelector("#" + el.getAttribute("aria-controls"))
+          let tabNextAfter = parseInt(divElement.getAttribute('data-nextafter'))
```
michitux, line 31: I find it more natural to use the [dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) property to retrieve values of data attributes, but that might just be a matter of taste. You could, e.g., name the attribute `data-next-after` and then retrieve it with `divElement.dataset.nextAfter`.

xwiki-pro-macros-api/src/main/resources/js/tabmacro.js
michitux: Okay, I understand the idea now, I think it would be good to add a comment what this `setTimeout` does (sleep) for those who don't know the pattern. Also, this for-loop doesn't feel natural. I wonder if it wouldn't be clearer to use a [do...while](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/do...while) loop with `loopCards === "true"` as condition. That should have the same effect but is much more explicit.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabMacro.java
```diff
@@ -0,0 +1,164 @@
...
+    {
+        // Don't show tab in edit mode.
+        Syntax syntax = context.getTransformationContext().getTargetSyntax();
+        SyntaxType targetSyntaxType = syntax == null ? null : syntax.getType();
+        if (SyntaxType.ANNOTATED_HTML.equals(targetSyntaxType) || SyntaxType.ANNOTATED_XHTML.equals(targetSyntaxType)) {
+            // Handle edit mode
+            ssrx.use("css/tabmacro.css");
+            List<Block> tabLabelBlock;
+            try {
+                Parser parser = componentManagerProvider.get().getInstance(Parser.class, "xwiki/2.1");
```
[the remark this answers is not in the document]
Josue-T: well, I just use a parser here because we can't just create a word block with the text as we need an other block type for special char. So in this case should we use the `plain/1.0` parser ?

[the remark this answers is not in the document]
Josue-T: @michitux, what do you think ?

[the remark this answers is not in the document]
michitux: Yes, otherwise, text like `**label**` will be rendered as `<strong>label</strong>`.

Issue #355, opened 2024-10-25
Create macro bridge for float macro


Josue-T:
This is to create a bridge macro of the confluence float macro: https://appfire.atlassian.net/wiki/spaces/CTFCSM/pages/471174205/Float

This will use the XWiki box macro: https://extensions.xwiki.org/xwiki/bin/view/Extension/Box%20Macro

PR #356, opened 2024-10-25
#355: Create macro bridge for float macro

Josue-T:
cf #355

### DIFF
--- a/xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceFloat.xml
+++ b/xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceFloat.xml
@@ -1,0 +1,621 @@
+<?xml version="1.1" encoding="UTF-8"?>
+
+<!--
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceFloat.xml
```diff
@@ -0,0 +1,621 @@
...
+        <name>type</name>
+        <number>5</number>
+        <prettyName>Parameter type</prettyName>
+        <size>60</size>
+        <unmodifiable>0</unmodifiable>
+        <classType>com.xpn.xwiki.objects.classes.StringClass</classType>
+      </type>
+    </class>
+    <property>
+      <defaultValue>40%</defaultValue>
```
trrenty, line 606: Should this be the default?

trrenty (approved):
Left a comment

xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceFloat.xml
Josue-T: Yes, the idea was to have a default value different than 0 as with 0 we might have some issues with the wysiwyg if the content is empty. And for 100% it don't really make sens as in this case it won't be really a float. So just took a value which make sens, it's better to have an other way to handle this.

Issue #323, opened 2024-08-14
Bulleted list does not work as expected when used in "Expand" macro


claudiuion117:
For the tests, I used Pro Macros 1.19.9. on XWiki 15.10.10., 15.10.11, 16.4.2. and 16.6.0. I also tested with Pro Macros 1.20.0. and got the same results.

Steps to reproduce:

- In a page, insert an "Expand" macro that has a Bulleted list in it, for example:

```
{{expand title="Click here to see the "Bulleted List" (not)."}}
The following test is a "Bulleted List"

        * Point number one

        * Point number two

Normal text

        * Point number three

        * Point number four

Normal text
{{/expand}}
```

- Click on Save & Exit
- Expand the macro
- Observe the page in View mode
- Click on Edit
- Observe the page in Edit mode

Expected results: The Bulleted list inside the macro is shown as a normal Bulleted list when the macro is expanded in both View and Edit mode

Actual results: In View mode, the Bulleted list has no indent and no bullet before the items. The order of the items and the content seems to be the same. However, in Edit mode, the list is shown as expected.

I think it is worth mentioning that using Numbered list instead of Bulleted list works as expected in both View and Edit modes: the lists are shown as a normal Numbered list when the section is expanded.

bullet list view mode:
![bullet-list-view-mode](https://github.com/user-attachments/assets/5c7923f4-bef5-4e02-9a05-064e52936081)

bullet list edit mode:
![bullet-list-edit-mode](https://github.com/user-attachments/assets/40f4e60e-88b4-483b-9469-45e233ce9509)

bullet vs number list in view mode:
![bullet-vs-number-lists](https://github.com/user-attachments/assets/c35fc488-d06f-4568-bb3e-1860bd7dd616)

number list view mode:
![number-list-view-mode](https://github.com/user-attachments/assets/01cc3aa7-5c83-440a-bd01-a4de08ffff02)

PR #357, opened 2024-10-25
The bulleted list does not work as expected when used in the "Expand"  macro (#323).

Farcasut:
* The issue was caused by a CSS rule on the .panel ul that removed the padding and the list-style-type. To fix this, I added a rule that targets the .confluence-expand-macro.panel.panel-default .panel-body ul and reverts the CSS style.


Farcasut:
image

trrenty (approved):

Issue #378, opened 2024-12-18
Tab group macro broken if we set an invalid ID


Josue-T:
If we put a content like this, it won't work because the ID `Hello World` is invalid.
```
{{tab-group id="Hello World"}}
{{tab label="Label 1"}}
content 1
{{/tab}}

{{tab label="label 2"}}
content 2
{{/tab}}
 
{{/tab-group}}
```

PR #379, opened 2024-12-18
#378: Tab group macro broken if we set an invalid ID

Josue-T:


trrenty (approved):
The code needs a little formatting

PR #388, opened 2025-01-14
Show-if/hide-if: Add display parameter

Josue-T:
Note that the current implementation is in Java, but an other solution would be to do it in CSS as described in this page: https://extensions.xwiki.org/xwiki/bin/view/Extension/PDF%20Export%20Application/#HAdaptthecontentforPDFexport

The java version is quite easier but have 2 inconvenient:
- In the HTML export, it's considered as printable which is not really correct.
- If we print directly from the browser (without using the PDF export tools), it's not considered as printable.

The PDF and ODT export work well.

### DIFF
--- a/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/internal/macro/AbstractShowHideIfMacro.java
+++ b/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/internal/macro/AbstractShowHideIfMacro.java
@@ -83,7 +83,8 @@ public abstract class AbstractShowHideIfMacro extends AbstractProMacro<ShowHideI
     {
         boolean matchAnyRes = false;
         boolean matchAllRes = true;
-        DocumentReference userReference = xwikiContextProvider.get().getUserReference();
+        XWikiContext xcontext = xwikiContextProvider.get();
+        DocumentReference userReference = xcontext.getUserReference();
         ShowHideIfMacroParameters.AuthType authTypeParam = parameters.getAuthenticationType();
         if (userReference != null) {
             UserReferenceList usersParam = parameters.getUsers();
@@ -124,6 +125,30 @@ public abstract class AbstractShowHideIfMacro extends AbstractProMacro<ShowHideI
                 break;
             default:
                 break;
+        }
+
+        if (parameters.getDisplayType() != null
+            && parameters.getDisplayType() != ShowHideIfMacroParameters.DisplayType.NONE)
+        {
+            boolean isExportPrintable = "export".equalsIgnoreCase(xcontext.getAction());
+            switch (parameters.getDisplayType()) {
+                case DEFAULT:
+                    if (isExportPrintable) {
+                        matchAllRes = false;
+                    } else {
+                        matchAnyRes = true;
+                    }
+                    break;
+                case PRINTABLE:
+                    if (isExportPrintable) {
+                        matchAnyRes = true;
+                    } else {
+                        matchAllRes = false;
+                    }
+                    break;
+                default:
+                    break;
+            }
         }
         return (parameters.getMatchUsing() == ShowHideIfMacroParameters.Matcher.ANY && matchAnyRes)
             || (parameters.getMatchUsing() == ShowHideIfMacroParameters.Matcher.ALL && matchAllRes);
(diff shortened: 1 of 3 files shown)


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/internal/macro/AbstractShowHideIfMacro.java
```diff
@@ -125,6 +126,30 @@ protected boolean doesMatch(ShowHideIfMacroParameters parameters)
...
+                    if (isExportPrintable) {
+                        matchAnyRes = true;
+                    } else {
+                        matchAllRes = false;
+                    }
+                    break;
+                default:
+                    break;
+            }
+        }
```
trrenty, line 152: the values of `matchAllRes` and `matchAnyRes` seem to be ovewritten. Is that intended?  This suggests that the values of the `authenticationType` and `groupReferenceList` params are ignored. Or that the they are taken in consideration in the following order 1.DisplayType 2.AuthType 3.GroupRefList. If so, this should be better specified in the parameter description so it will be clear for the users.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/internal/macro/AbstractShowHideIfMacro.java
```diff
@@ -125,6 +126,30 @@ protected boolean doesMatch(ShowHideIfMacroParameters parameters)
...
+        if (parameters.getDisplayType() != null
+            && parameters.getDisplayType() != ShowHideIfMacroParameters.DisplayType.NONE)
+        {
+            boolean isExportPrintable = "export".equalsIgnoreCase(xcontext.getAction());
+            switch (parameters.getDisplayType()) {
+                case DEFAULT:
+                    if (isExportPrintable) {
+                        matchAllRes = false;
+                    } else {
+                        matchAnyRes = true;
```
trrenty, line 139: Shouldnt it be something like this?

```suggestion
                        matchAllRes = false;
                        matchAnyRes = false;
                    } else {
                        matchAllRes = true;
                        matchAnyRes = true;
```

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/macro/ShowHideIfMacroParameters.java
```diff
@@ -147,4 +168,20 @@ public void setGroups(GroupReferenceList groups)
     {
         this.groups = groups;
     }
+
+    /**
+     * @return the display type which should match.
```
trrenty, line 173: the description is a bit vague. From what I understand, this parameter should add some additional displaying constraints? If it's NONE, this parameter is ignored and the macro will display according to the other params. DEFAULT will display in all the context actions (view/edit/etc) except `export` and PRINTABLE will display only when the context action is export?

You should also add the @PropertyDescription annotation. Ideally, some translations should be added.

trrenty (changes requested):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/macro/ShowHideIfMacroParameters.java
trrenty: This is how you create translations for the macro parameters https://www.xwiki.org/xwiki/bin/view/Documentation/DevGuide/Tutorials/WritingMacros/WikiMacroTutorial/#HTranslations

The current implementation is alright. We can improve it in the future if needed be.

trrenty (approved):

Josue-T:
@trrenty well, so I discover now that using `RenderingScriptService` break the build with this error:
```
You must inject a component role. Got [org.xwiki.rendering.script.RenderingScriptService] at (/home/josue/Git/xwiki-sas/xwiki-pro-macros/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/internal/macro/AbstractShowHideIfMacro.java:75)
```
Any idea how to solve it ?

trrenty:
Lets continue the conversation here https://github.com/xwikisas/xwiki-pro-macros/pull/389/files#r1916811736

PR #389, opened 2025-01-14
Show-if/hide-if: Show error if the somes parameter are unsupported

Josue-T:


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/internal/macro/AbstractShowHideIfMacro.java
```diff
@@ -128,4 +138,36 @@ protected boolean doesMatch(ShowHideIfMacroParameters parameters)
...
+            if (!beanPropertiesIds.contains(parameterName) && !parametersWhiteList.contains(parameterName)) {
+                unsupportedParameters.add(parameterName);
+            }
+        }
+        if (!unsupportedParameters.isEmpty()) {
+            return Optional.of(
+                new MacroBlock("error", Collections.emptyMap(),
+                    // TODO escape content but seem that XWikiSyntaxEscaper
+                    //  is only available since XWiki-platform 14.10.6
+                    "Unsupported parameter for macro: " + String.join(", ", unsupportedParameters), false));
```
Josue-T, line 169: @trrenty  do you have any idea how to manage escaping for a macro content when we are on XWiki 14.10.0 ?

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/internal/macro/AbstractShowHideIfMacro.java
```diff
@@ -128,4 +138,36 @@ protected boolean doesMatch(ShowHideIfMacroParameters parameters)
...
+     * confluence and on confluence a lot more parameter are supported. In case of the parameter is not supported it
+     * could confuse the user thy this macro don't work as expected.
+     *
+     * @param context
+     * @return
+     */
+    protected Optional<Block> maybeGetUnsupportedParameterErrorBlock(MacroTransformationContext context)
+    {
+        List<String> parametersWhiteList = List.of("atlassian-macro-output-type");
+        java.util.Map<String, String> allParameters = context.getCurrentMacroBlock().getParameters();
```
trrenty, line 153:
```suggestion
        Map<String, String> allParameters = context.getCurrentMacroBlock().getParameters();
```
why not import it

trrenty (changes requested):

trrenty (approved):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/showhideif/internal/macro/AbstractShowHideIfMacro.java
```diff
@@ -59,6 +71,9 @@ public abstract class AbstractShowHideIfMacro extends AbstractProMacro<ShowHideI
     @Inject
     private GroupManager groupManager;
 
+    @Inject
+    private RenderingScriptService renderingScriptService;
```
trrenty, line 75: @Josue-T 

Should have been
```
    @Inject
    @Named("rendering")
    private ScriptService renderingScriptService;
```
And then cast it to RenderingScriptService when using it.

It is hackish, but I'm not sure we have other solutions right now.

You should also add a TODO comment to make use of `XWikiSyntaxEscaper` after upgrading the parent of the app to something greater than 14.10.6

Josue-T: Thank you a lot @trrenty, if work this way.

PR #422, opened 2025-01-21
make Profile picture macro support inline mode #414

ChiuchiuSorin:
Adapted the team macro to be used inline, as the Profile picture macro is based on the team macro.


oanalavinia (approved):
Looks good overall

Issue #391, opened 2025-01-16
Excerpt-include macro don't detect except macro when except macro is into an other macro


Josue-T:
Step to reproduce:

1. Create a page with this content (at this reference: `Sandbox.exerptsource.WebHome`):
```
{{confluence_layout}}
{{confluence_layout-section ac:type="single"}}
{{confluence_layout-cell}}
{{excerpt name="name" inline="true"}}
my content from except
{{/excerpt}}
{{/confluence_layout-cell}}
{{/confluence_layout-section}}
{{/confluence_layout}}
```
2. Create an other page with this content
```
{{excerpt-include 0="Sandbox.exerptsource.WebHome" name="name" nopanel="true"/}}
```

Expected: we see the content

Currently: we have this error:

![Image](https://github.com/user-attachments/assets/da6cb9d8-8171-438e-a444-5fc47da9ebd8)

The reason of this issue is that we try to do a search in the XDOM before we do the rendering of the macro content and so we get only the top level macro and not the inner macro.

ClemensRobbenhaar:
PR at https://github.com/xwikisas/xwiki-pro-macros/pull/441

PR #441, opened 2025-01-25
Excerpt-include macro don't detect except macro when except macro is …

ClemensRobbenhaar:
…into an other macro #391

* if a macro is not the excerpt macro, but has content which is wiki syntax, descent into the macro content to look for excepts there (this implements a "depth first" search, which loosely works like a "first matching macro in wiki text" search)
* to recurse into the macro content write yet another helper function and inject / look up the necessary components to parse the macro content into a XDOM

The new helper function `getExcerptFromXDOM` takes an inordinate number of parameters. Maybe introducing a "parameter object" could help to reduce these. I can update the PR to introduce  this, if wanted.


trrenty (approved):
Looks good

Issue #312, opened 2024-07-27
When added in a page, the view file macro proposes the deprecated att--filename parameter as the first parameter to fill in, while name is displayed in the "More" parameters


lucaa:
How to reproduce:
* add a view-file macro in a page, from the rich text editor
* fill in the parameters

Expected result:
* the name parameter, to name the file to display, is displayed as the first parameter (and mandatory ?)
* the compatibility-only parameter att--filename is less visible (or not visible at all)

Actual result:
* the att-filename parameter is the first one displayed
![image](https://github.com/user-attachments/assets/8f79e062-7833-4353-873e-84c843fcd5d9)
* the name parameter is not even displayed on the first screen, it's displayed under the "More" button
![image](https://github.com/user-attachments/assets/1b41d4ae-eab2-4141-bdd8-c8f5a5ee9be7)

lucaa:
tested with Pro macros 1.19.9 on XWiki 16.4.1

snazare:
done as part of #450

Issue #307, opened 2024-07-27
View file macro display parameter description says the default is full when it actually seems to be thumbnail


lucaa:
In the macro parameters, the display parameter says the following:

![image](https://github.com/user-attachments/assets/df7fffb6-8c75-4542-86a6-2637c64fd4cc)

However, after adding the macro with no value filled in for the display parameter, the display looks more like what "thumbnail" would display:

![image](https://github.com/user-attachments/assets/c154a03c-2bb9-41e4-810c-2cf51ba65db9)

I think thumbnail is a good default, we should just make sure that the documentation is up to date.

lucaa:
tested with Pro macros 1.19.9 on XWiki 16.4.1

snazare:
done as part of #377

ChiuchiuSorin:
Set the thumbnail as default and updated the hint.

Issue #306, opened 2024-07-27
View file macro's name is uselessly misleading, using the plural


lucaa:
The macro appears in the macros list and in the  slash action suggests as "View files", it's not clear why the plural "files" since the id of the macro is singular (`view-file`) and the macro only allows to view a single file, so I don't understand why it appears in plural.

![image](https://github.com/user-attachments/assets/2fd1978d-1263-43b5-8e2d-a54cbfbaaa38)

As a page editor reading that list, I expect see there a description of the action that I will do should I choose that option, which in case it's "view file", even if, as a general description, the macro allows to "view files" (because it will be used multiple times and "view file" will happen multiple times).

lucaa:
Looking at the other macros, I don't see many that use the plural like this...

lucaa:
tested with Pro macros 1.19.9 on XWiki 16.4.1

snazare:
done as part of #377

PR #458, opened 2025-01-30
Add a page parameter, that will dictate the attachments of which page are being displayed in the attachment picker #454

ChiuchiuSorin:
Moved the view file macro to java:
* added the macro parameters class
* moved the velocity logic to a template and adapted it
* added JavaScript code to correlate the page parameter to the attachment parameter
* added translations
* removed leftover enum

Fixed issues: 
* #454
* #450
* #312
* #307
* #306


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/ViewFile.xml
```diff
@@ -178,7 +178,77 @@ Full Presentation:
...
+      <code>require.config({
+  paths: {
+    'xwiki-suggestAttachments': "$xwiki.getSkinFile('uicomponents/suggest/suggestAttachments.js', true)" +
+    "?v=$escapetool.url($xwiki.version)"
+  }
+});
+
+require(['jquery', 'xwiki-suggestAttachments'], function($) {
+  const acceptedExtensions = '.ppt,.pptx,.odp,.doc,.docx,.odt,.xls,.xlsx,.ods,.pdf';
+  const styleObserver = new MutationObserver(function(mutations) {
```
oanalavinia, line 190: I'd add a comment to explain why is this needed, here or for other parts that are not so obvious. It makes sense in the context of this PR why all changes are needed, but it can be harder when someone tries to understand the reason. It will also help to mention that this `trick` will be removed once there is a proper way to do it in the platform (and link a jira issue)

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Translations.xml
```diff
@@ -234,10 +234,15 @@ com.xwiki.macros.internal.recentlyupdated.RecentlyUpdatedThemeEnum.social=User,
...
 ## View Files macro
-com.xwiki.macros.viewfile.DisplayType.full=Full
-com.xwiki.macros.viewfile.DisplayType.thumbnail=Thumbnail
+rendering.macro.view-file.description=Show a file using PDF Viewer Macro or Office Viewer.
+rendering.macro.view-file.parameter.attFilename.name=Att--filename
+rendering.macro.view-file.parameter.attFilename.description=Alias of File name (here for compatibility reasons).
+rendering.macro.view-file.parameter.name.name=File name
+rendering.macro.view-file.parameter.name.description=The attachment reference to display.
+rendering.macro.view-file.parameter.page.name=Page
+rendering.macro.view-file.parameter.page.description=The source page of the attachment.
```
oanalavinia, line 243: I would add more details to explain that this is needed for the file param and changing it will reset the file value

oanalavinia (changes requested):

oanalavinia:
Please also add the title + id of the issues fixed in this PR in the description. Even if maybe they are not needed in the description (because are linked using github), you will need to mention each issue in the merge commit (so to have a proper git history)

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/macro/ViewFileMacro.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Inject
+    private ScriptContextManager scriptContextManager;
+
+    /**
+     * Create and initialize the descriptor of the macro.
+     */
+    public ViewFileMacro()
+    {
+        super("View file", "Show a file using PDF Viewer Macro or Office Viewer.", ViewFileMacroParameters.class);
```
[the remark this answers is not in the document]
ChiuchiuSorin: They are default values, the translations overwrite them.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/macro/ViewFileMacro.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public boolean supportsInlineMode()
+    {
+        return true;
```
trrenty, line 67: I believe that only a bunch of view modes support inline? If so, it would be nice to display an error depending on the the display type and whether it's inline or not. In order to to avoid display issues.

For example, display "FULL" should not be displayed inline, but if the user does, it's really probable that the ckeditor content will be messed up.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/macro/ViewFileMacro.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        Template customTemplate = this.templateManager.getTemplate("viewfile/viewFileTemplate.vm");
+        ScriptContext scriptContext = scriptContextManager.getScriptContext();
+
+        scriptContext.setAttribute("params", parameters, ScriptContext.ENGINE_SCOPE);
+        scriptContext.setAttribute("isInline", context.isInline(), ScriptContext.ENGINE_SCOPE);
+        try {
+            return this.templateManager.execute(customTemplate).getChildren();
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
trrenty, line 82: I think a `MacroExecutionException` is more appropriate.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/macro/ViewFileMacroParameters.java
```diff
@@ -0,0 +1,164 @@
...
+    @PropertyDisplayType(DocumentReference.class)
+    public void setPage(String page)
+    {
+        this.page = page;
+    }
+
+    /**
+     * @return the display type.
+     */
+    public ViewFileDisplay getDisplay()
```
trrenty, line 92: You shouldn't expose internal classes in public apis. You could make `ViewFileDisplay` a part of the parameters.

trrenty: You can also set the `@PropertyDisplayType` to ViewFileDisplay in order to have a nice picker for it.

trrenty (changes requested):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/macro/ViewFileMacro.java
ChiuchiuSorin: The way it works is that when used inline, the thumbnail option is forced even if the full display is selected. This is behavior was added before my changes.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/macro/ViewFileMacroParameters.java
ChiuchiuSorin: > You can also set the `@PropertyDisplayType` to ViewFileDisplay in order to have a nice picker for it.

There is no need as the field type declaration is enough to add the picker.

oanalavinia (approved):
Looks good :) 
I think there are only the question from Teo

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/macro/ViewFileMacro.java
trrenty: I feel like this might cause more confusion, but we will see whether people complain or not.

PR #461, opened 2025-01-30
Make the title field editable inplace too #399

ChiuchiuSorin:
* added a check for inplace and wysiwyg editing and replaced the HTML structure with velocity in these cases to avoid errors related to wikimacroparameter and wikimacrocontext
* adapted CSS
* added default title value
Demo:

https://github.com/user-attachments/assets/39b09231-eae6-488c-a051-ec0eb0e0fda5

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
@@ -503,7 +503,7 @@ Hello 👀
       <contentJavaType>Wiki</contentJavaType>
     </property>
     <property>
-      <contentType>Optional</contentType>
+      <contentType>Mandatory</contentType>
     </property>
     <property>
       <defaultCategories/>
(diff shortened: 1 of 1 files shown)


ChiuchiuSorin:
To avoid an issue regarding `{{wikimacroparameter}}` and `{{wikimacrocontent}}` that fail to render when in edit mode while inside a HTML macro, we need to manually write the HTML by using velocity syntax. This way, while the macro will always be expanded when in edit mode, the user will be able to modify the text inline.

snazare:
HI @ChiuchiuSorin I agree with the proposed solution, as it is the most common usecase for the end user to intend to edit both the title and content. Thank you very much!

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
```diff
@@ -429,7 +433,29 @@ Hello 👀
     <property>
       <code>{{velocity output="false"}}
 #macro (executeMacro)
+  ## To avoid an issue regarding {{wikimacroparameter}} and {{wikimacrocontent}} that fail to render when in
+  ## edit mode while inside a HTML macro, we need to manually write the HTML by using velocity syntax. This way,
```
Farcasut, line 437:
```suggestion
  ## edit mode while inside a HTML macro, we need to manually write the HTML by using xwiki syntax. This way,
```
I think you have a typo because the syntax below is XWiki syntax.

Farcasut (approved):

Issue #430, opened 2025-01-23
Update the description of the type parameter


Farcasut:
To "Select one or multiple content types. The macro will display only the content of the type you specify here." Note: please keep the support for the confluence parameters so that we continue to support migration of this macro. Please include the current description of the parameter in the documentation.

PR #462, opened 2025-01-30
Update the description of the type parameter #430 Make the type parameter a static list picker #429

Farcasut:
Updated the type parameter to use a picker and added a new translation. The code that was handling Confluence-like parameters is still present, so the migrations are not affected.


xwiki-pro-macros-api/src/main/resources/templates/html_displayer/recentlyupdatetype/edit.vm
```diff
@@ -0,0 +1,33 @@
...
+##
+## You should have received a copy of the GNU Lesser General Public
+## License along with this software; if not, write to the Free
+## Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+## 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+## ---------------------------------------------------------------------------
+#template("pickers_macros.vm")
+#set ($parameters = {
+  'name': $displayer.parameters.name,
+  'value': '',
```
trrenty, line 23: i think this causes a bug. If you open the modal a second time (after you placed some values), the inputs will be empty.

trrenty (changes requested):

PR #464, opened 2025-01-31
Make the spaces parameter a page picker with multiple choices #432 & Update the description of the space parameter #438

Farcasut:
Added a picker for the space parameter. However, since the space parameter had more responsibilities when coming from Confluence, I added three additional parameters:

1.Excluded space: Handles excluded spaces, ensuring that documents from them are ignored, similar to how -space worked before.
2.Global: Functions the same as the old @all, where the search is performed across the current wiki and its subwikis.
3.Current Wiki: Behaves like "Global," but the search is restricted to the current wiki only.

I kept the old default value of the space parameter. As for precedence, the order is as follows: Global, Current Wiki, and Space.

Uploading 2025-01-31 15-29-30.mp4…


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdated.xml
```diff
@@ -865,6 +867,39 @@ require(['jquery'], $ =&gt; {
...
+  #if (!$confluenceSpaces.trim().isEmpty() &amp;&amp; !$exclude.trim().isEmpty())
+    #set ($finalSpace = "${confluenceSpaces},${exclude}")
+  #elseif(!$confluenceSpaces.trim().isEmpty())
+    #set ($finalSpace = $confluenceSpaces)
+  #else
+    #set ($finalSpace = $exclude)
+  #end
+#end
+
+## Determinate if the search should be done in the currentWiki, globally or only on specific spaces.
```
trrenty, line 892:
```suggestion
## Determine if the search should be done in the currentWiki, globally or only on specific spaces.
```

trrenty (approved):

Issue #305, opened 2024-07-27
Adding a view-files macro before a block macro in the rich text editor adds its result as content in the page


lucaa:
How to reproduce:
* on a page, add a block macro, e.g. info:
```
{{info}}test{{/info}}
```
* open this page in the rich text editor
* position the caret before the info box (using the red newline button of the editor to create a new line)
* insert a view-file macro, either using the quick actions menu or using the insert macro menu
* select or not a file to view, depending on how the macro was added it may or may not be asked for
* save page

Expected result:
* the content of the page is made of 2 macro calls, a view-file macro call and the info macro

Actual result
* there are the 2 macro calls in the page content, but there is also some content as wiki syntax between the 2, which represents the display of the view file macro:
```
{{view-file/}}

(% class="box errormessage" %)
(((
rendering.macro.viewFile.namerequired"
)))

{{html}}
<span class="viewFileThumbnail viewFileThumbnailButton"  data-preview="false" data-ref="xwiki:Sandbox.TestPage1@filename">
<a href="/xwiki/bin/download/Sandbox/TestPage1/filename"
download="download"
 class="button button-primary" title="View file content"
>
  <span class="attachmentMimeType" data-type="">
                                                                                  <span title="Attachment"><span class="fa fa-paperclip" aria-hidden="true"></span></span>
      </span>

<span class="viewFileName">filename</span>
</a>
</span>
{{/html}}

{{info}}
test
{{/info}}
```
Note: it happens the same if I fill in an actual file in the view-file macro.

Reproduced on Chrome, on XWiki 16.4.1, may be a new issue related to the version of XWiki.

lucaa:
Note: this reproduces also if I add a view-file macro before another view-file macro.

lucaa:
tested with Pro macros 1.19.9 on XWiki 16.4.1

snazare:
done as part of #377

PR #469, opened 2025-02-03
Adding a view-files macro before a block macro in the rich text editor adds its result as content in the page #305

ChiuchiuSorin:
When using the view file macro inline, the `CKEditor` will render` the error macro` outside of the original macro, generating additional HTML. To avoid this, I added a manual check for when the macro is being used inline and generated the appropriate HTML structure manually.


xwiki-pro-macros-api/src/main/resources/templates/viewfile/viewFileTemplate.vm
```diff
@@ -111,10 +111,21 @@
...
   #if (!$unescapedFilename)
-
-    {{error}}
-       $escapetool.xml($services.localization.render('rendering.macro.viewFile.attachmentrequired'))
-    {{/error}}
+    ## When using the error macro inline in another macro, the CKEditor will render it outside of the original macro,
+    ## generating additional HTML. To avoid this, we manually identify when the macro is being used inline and generate
+    ## the appropriate HTML structure.
+    #if ($isInline)
+      {{html clean='false'}}
```
trrenty, line 118: is "clean=false" needed here?

trrenty (approved):

xwiki-pro-macros-api/src/main/resources/templates/viewfile/viewFileTemplate.vm
ChiuchiuSorin: Yes. Removing `clean='false'` will cause the editor to automatically try to convert to XWiki Syntax which generates a `<p>` element. This, when used inline will break the display (as the editor will take the block element out of the inline macro).

Issue #496, opened 2025-03-17
View File macro: file name out of bounds for long strings without dashes or spaces


lamblinc:
For files with long names - long string, no dashes, no space in it - using view files in thumbnail mode ends-up with a name displayed out of expected bounds, leading to overlapping content (for macros in contiguous table cells for example).

1. Upload an attachment to a page with long file name and no dashes in it (example A1479724_EDITION.BC_SIGMABC_662156_0001_001_11460284.pdf)
2. Add a view file macro in this page with default options, linked to the attachment.

Expected behavior:
The file name is wrapped nicely under the icon of the file, respecting its width.

Screenshot:

![Image](https://github.com/user-attachments/assets/360f130a-746c-44ce-8274-77840d5bccc3)

 - OS: Lnux (Arch based)
 - Browser : FF 136, Chromium 134.0.6998.88
 - XWiki version: 16.4.6, 16.10.2
 - App version : 1.26.7, 1.26.4 at least

Additional context:
Text break seems only to be possible with dashes or space - dots and underscore do not work.

PR #497, opened 2025-03-18
View File macro: file name out of bounds for long strings without dashes or spaces #496

Farcasut:


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/ViewFile.xml
```diff
@@ -438,10 +438,10 @@ window.addEventListener("DOMContentLoaded", function () {
 
 .viewFileThumbnailCard, .viewFileFull {
   position: relative;
-  width: 10rem;
+  width: 10rem !important;
```
Farcasut, line 441: These !important are needed to override the  wysiwyg editor which adds width: min-content. The line breaks added by overflow-wrap: break-word are ignored when the minimum content width is calculated. As a result, in edit mode, the preview will expand horizontally if the text is really long.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/ViewFile.xml
```diff
@@ -467,10 +467,15 @@ span.viewFileThumbnailCard {
   width: 100%;
   border: 1px solid;
   border-radius: 1rem;
+  min-height: 60px;
```
Farcasut, line 470: We need this min-height to ensure that the thumbnail doesn't get squashed when the title is really long and displayed is displayed on multiple lines

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/ViewFile.xml
```diff
@@ -467,10 +467,15 @@ span.viewFileThumbnailCard {
...
 }
 
 .viewFileModal-downloadLink {
   margin-left: 2px;
+}
+
+.viewFileName {
```
Farcasut, line 477: The rule that actually adds the line breaks if the title overflows

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/ViewFile.xml
oanalavinia, line 470: Could you maybe use `em` instead of `px`?

oanalavinia (approved):
Added just one comment

Issue #583, opened 2025-07-14
Integration tests of macros defined in xwiki pages, that have parameters with custom types, fail to render.


trrenty:
Currently, the docker tests module only tests the Team macro. This macro has a parameter with a custom `com.xwiki.pickers.UsersReference` type.

If the type of the parameter is blank/or a more common type i.e. Date/String/Integer the test succeeds.

If the test is run with an external instance, it runs.

A POSSIBLE SOLUTION that needs testing is to set the custom type to a common class that is parameterized. i.e. replace `com.xwiki.pickers.UsersReference` with `java.util.List(com.xwiki.pickers.UsersReference)`.

A possible cause to the problem is that some dependency is missing. This suspicion is due to the fact that the test succeeds on a external rerun. (run the test normally -> fail; run the test with the created instance -> succeed). Intuitively, the issue should reproduce on a fresh 14.10 instance, but it doesn't. This suggests that on a fresh flavored xwiki instance, there is some code that gets executed. 

The issue might not reproduce on higher xwiki versions.

If everything else fails, i think the macros might have to be converted into java macros or some workaround should be found.

Farcasut:
> A POSSIBLE SOLUTION that needs testing is to set the custom type to a common class that is parameterized. i.e. replace com.xwiki.pickers.UsersReference with java.util.List(com.xwiki.pickers.UsersReference).

This may work, but it will also break the picker because we will be using the list picker instead of the custom one.

PR #584, opened 2025-07-16
Integration tests of macros defined in xwiki pages, that have parameters with custom types, fail to render #583

Farcasut:
The problem was that the macros were not registered when the app was installed, but resaving the page fixes the issue. For the time being, this is the only solution I could find(Thanks to Marius).


trrenty (approved):
Good job on finding a workaround

Issue #266, opened 2024-06-03
View file macro: Gracefully handle big attachments


raphj:
Currently, big attachments can cause heap space errors and out of memory exceptions, killing wikis.

Big attachments should be gracefully handled.

Ideas:
1. stop after a given amount of generated XDOM or words
2. automatically user the new thumbnail display for attachments over a given size
3. make the thumbnail display the default

I believe all of these things should be implemented.

See also #267

ChiuchiuSorin:
Could you give me some metrics as to what would be considered a big attachment? Or in which conditions did this issue occur?

lucaa:
@raphj while I understand the point of this issue, I think the display of the view-file macro with display=full should be identical to the display of the office macro for the same file, see #309.

It appears that the full display mode is no longer default (if it ever was) - see #307 - so to me it's somewhat reasonable that if the parameter is set explicitly to full to actually respect it, and leave it to the user to decide which for which attachment full may be too much. We'd just need to catch exceptions and display properly that the doc could not be rendered, should that happen. This also saves us from deciding what's big and what's not. We can add a warning about resource consumption in the description of the display mode parameter.

ChiuchiuSorin:
After discussing internally, the solution will be not to render large files and only make them available for download.

After discussing internally, we came with two implementation options:

* add a paramter in configuration for view file macro to limit the size of a file - or set a hard limit in relation with the instance allocated memory (like a known instance limitation)
* get available memory and set a minimum threshold of available memory. If the instance has less then the minimum threshold and the file is of a specific size (set some size intervals), apply solution - this may require more time as it will involve more testing to get the right ratio between available memory and file size.

PR #588, opened 2025-07-21
View file macro: Gracefully handle big attachments #266

ChiuchiuSorin:
### Root issue:
The root cause of the issue comes from the office macro, I'm assuming, from the converter and the applied image filters: https://github.com/xwiki/xwiki-platform/blob/2baba8175f80b2b915db5b88bba988b24cff89fe/xwiki-platform-core/xwiki-platform-office/xwiki-platform-office-importer/src/main/java/org/xwiki/officeimporter/internal/server/DefaultOfficeServer.java#L151-L153
I think an easy fix for this issue would be adding a `taskExecutionTimeout` to the conversion builder. 
The error that I got when the rendering failed:
```
2025-07-18 16:11:04,436 [jodconverter-poolentry-1] ERROR o.j.l.t.LocalConversionTask    - Local conversion failed. 
org.jodconverter.core.office.OfficeException: Could not apply filter org.jodconverter.local.filter.text.LinkedImagesEmbedderFilter.
 at org.jodconverter.local.filter.AbstractFilterChain.doFilter(AbstractFilterChain.java:114)
 at org.jodconverter.local.filter.AbstractFilterChain.doFilter(AbstractFilterChain.java:91)
 at org.jodconverter.local.filter.DefaultFilterChain.doFilter(DefaultFilterChain.java:98)
 at org.jodconverter.local.task.LocalConversionTask.modifyDocument(LocalConversionTask.java:167)
 at org.jodconverter.local.task.LocalConversionTask.execute(LocalConversionTask.java:118)
 at org.jodconverter.local.office.LocalOfficeManagerPoolEntry.doExecute(LocalOfficeManagerPoolEntry.java:120)
 at org.jodconverter.core.office.AbstractOfficeManagerPoolEntry.lambda$execute$0(AbstractOfficeManagerPoolEntry.java:80)
 at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
 at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
 at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
 at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: com.sun.star.lang.DisposedException: null
 at com.sun.star.lib.uno.environments.remote.JobQueue.removeJob(JobQueue.java:210)
 at com.sun.star.lib.uno.environments.remote.JobQueue.enter(JobQueue.java:330)
 at com.sun.star.lib.uno.environments.remote.JobQueue.enter(JobQueue.java:303)
 at com.sun.star.lib.uno.environments.remote.JavaThreadPool.enter(JavaThreadPool.java:87)
 at com.sun.star.lib.uno.bridges.java_remote.java_remote_bridge.sendRequest(java_remote_bridge.java:636)
 at com.sun.star.lib.uno.bridges.java_remote.ProxyFactory$Handler.request(ProxyFactory.java:146)
 at com.sun.star.lib.uno.bridges.java_remote.ProxyFactory$Handler.invoke(ProxyFactory.java:128)
 at com.sun.proxy.$Proxy178.getGraphicObjects(Unknown Source)
 at org.jodconverter.local.filter.text.LinkedImagesEmbedderFilter.convertLinkedImagesToEmbedded(LinkedImagesEmbedderFilter.java:85)
 at org.jodconverter.local.filter.text.LinkedImagesEmbedderFilter.doFilter(LinkedImagesEmbedderFilter.java:68)
 at org.jodconverter.local.filter.AbstractFilterChain.doFilter(AbstractFilterChain.java:110)
 ... 10 common frames omitted
Caused by: java.io.IOException: com.sun.star.io.IOException: EOF reached - socket,host=127.0.0.1,port=8100,tcpNoDelay=1,localHost=localhost,localPort=51596,peerHost=localhost,peerPort=8100
 at com.sun.star.lib.uno.bridges.java_remote.XConnectionInputStream_Adapter.read(XConnectionInputStream_Adapter.java:55)
 at java.base/java.io.DataInputStream.readInt(DataInputStream.java:392)
 at com.sun.star.lib.uno.protocols.urp.urp.readBlock(urp.java:355)
 at com.sun.star.lib.uno.protocols.urp.urp.readMessage(urp.java:92)
 at com.sun.star.lib.uno.bridges.java_remote.java_remote_bridge$MessageDispatcher.run(java_remote_bridge.java:105)
```
### Applied solutions:

- Added thumbnail conversion timeout.
- Added a new class to validate the size of an attachment with checks:
  - for the number of paragraphs of a DOCX/ODT file .
  - that the zip entry size of a `Microsoft open XML` file or `open document` file formats are above the Apache POI max record size (found it to be 100,000,000).
- Files that are oversize, are forced to thumbnail (no image) and can only be downloaded.

I've found from error logs (both from cloud and local instances) that the Apache POI max record size is 100,000,000:
```
Caused by: org.apache.poi.util.RecordFormatException: Tried to read data but the maximum length for this record type is 100,000,000.
If the file is not corrupt or large, please open an issue on bugzilla to request 
increasing the maximum allowable size for this record type.
```
**This solution only applies to**  `Microsoft open XML` file or `open document` file formats, not to the old `Microsoft Word Binary File format` (.doc, .xls, .ppt files). A solution for those files could be imposing a hard limit based on the worst case scenario. I'm not sure this would be a great solution, as I tested .doc files of over 30mb that contained images and they were successfully rendered, but failed to render a 17mb file that had over 200k paragraphs. This, in my opinion, is a short, partial fix until the root problem is fixed in the platform. 

### Other methods that were tried to fix the issue:

1. Attempted to find a relation:
    - Between the memory available and the file size to try prevent the rendering of an attachment if the memory is too low.
    - Between error occurrence and the file size, to limit the rendering of files above a specific size.
    - Failed: while attempting to find a relation by trial and error, I've noticed that the size of the file or the available memory are not actually the main concern. Zip files (Microsoft 2007 and Open document formats) that were larger then 14mb were successfully rendered, while files smaller then 1mb failed. This was caused due to the fact that the smaller files had massive amount of text and paragraphs, while the larger ones also contained images which were easier to render.
    - I did notice a relationship between large amounts of paragraphs and OOM errors. From testing, I could render files with up to 200k paragraphs, but above 230k it would fail the rendering, no matter the allocated memory (tried to allocate up to 3gb of memory). On the other hand, documents with a lot of characters (tested up to 7m) would not fail to render as long as they are split in paragraphs (one massive paragraph with a few milion characters would easily break the instance). Therefor, I opted for a check for the number of paragraphs.

2. Attempted to create a mock conversion with a set execution timeout
    - Failed: While the idea of a mock conversion of the target file actually succeeded in excluding very large files because the execution timeout was reached, this added a lot of additional processing time to medium files that now took a few seconds to check if they can be converted (with the mock conversion) and then another few seconds to actually be rendered by the office macro.

3. Attempted to accurately read the number of paragraphs in a document by opening it with Apache POI.
    - Failed: To accurately read the number of paragraphs in a document, I'd have to open the document which sometimes breaks the instance (usually when going from inline edit to view mode).
    - Solution: open the `DOCX` or `ODT` as zip files and check the metadata to find the number of paragraphs. While this works, it is however limited by the need that the editor should update the metadata every time the document is edited. Moreover, this cannot be applied to `doc` files, as those are binary files.

### DIFF
--- a/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/AttachmentSizeValidator.java
+++ b/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/AttachmentSizeValidator.java
@@ -1,0 +1,171 @@
+/*
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation; either version 2.1 of
+ * the License, or (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this software; if not, write to the Free
+ * Software Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
+ * 02110-1301 USA, or see the FSF site: http://www.fsf.org.
+ */
+package com.xwiki.macros.viewfile.internal;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import javax.inject.Inject;
+import javax.inject.Provider;
+import javax.inject.Singleton;
+
+import org.apache.commons.lang.exception.ExceptionUtils;
+import org.apache.poi.util.IOUtils;
+import org.slf4j.Logger;
+import org.xwiki.component.annotation.Component;
+import org.xwiki.model.reference.AttachmentReference;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.XWikiAttachment;
+import com.xpn.xwiki.doc.XWikiDocument;
+
+/**
+ * Checks if Office 2007 and ODF document attachments are oversize by checking the ZIP entries size or if an excessive
+ * amount of paragraphs is present. This validator prevents documents from causing performance or memory issues during
+ * processing (e.g., conversion or rendering), by applying constraints derived from Apache POI and LibreOffice
+ * limitations.
+ *
+ * @version $Id$
+ * @since 1.27.1
+ */
+@Component(roles = AttachmentSizeValidator.class)
+@Singleton
+public class AttachmentSizeValidator
+{
+    private static final int MAX_ENTRY_LENGTH = 100_000_000;
+
+    private static final List<String> OLD_OFFICE_EXTENSIONS = List.of("doc", "xls", "ppt");
+
+    private static final int MAX_PARAGRAPHS = 200_000;
+
+    @Inject
+    private Provider<XWikiContext> wikiContextProvider;
+
+    @Inject
+    private Logger logger;
+
+    /**
+     * Determines whether a given document attachment exceeds safe limits for processing. This method only applies to
+     * Office 2007 and ODF formats attachments.
+     *
+     * @param attachRef The reference to the attachment to check
+     * @return {@code true} if the attachment is oversize, or {@code false} otherwise
+     * @throws XWikiException if an error occurs while attempting to retrieve the attachment and it's content
+     * @throws IOException if an error occurs while reading or parsing the attachment
+     */
+    public boolean isAttachmentOversize(AttachmentReference attachRef) throws XWikiException, IOException
+    {
+        XWikiContext wikiContext = wikiContextProvider.get();
+        XWikiDocument document = wikiContext.getWiki().getDocument(attachRef.getDocumentReference(), wikiContext);
+        XWikiAttachment attachment = document.getAttachment(attachRef.getName());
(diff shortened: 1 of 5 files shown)


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/AttachmentSizeValidator.java
```diff
@@ -0,0 +1,171 @@
...
+import org.xwiki.component.annotation.Component;
+import org.xwiki.model.reference.AttachmentReference;
+
+import com.xpn.xwiki.XWikiContext;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.XWikiAttachment;
+import com.xpn.xwiki.doc.XWikiDocument;
+
+/**
+ * Checks if Office 2007 and ODF document attachments are oversize by checking the ZIP entries size or if an excessive
```
oanalavinia, line 48: I guess this class doesn't make much sense after the XWiki issue is fixed, right? So I would add a comment to remove this extra check once that issue is fixed (and the parent is upgraded)

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/AttachmentSizeValidator.java
```diff
@@ -0,0 +1,171 @@
...
+
+    private boolean containsOversizeEntry(InputStream file) throws IOException
+    {
+        int maxEntrySize = IOUtils.getMaxByteArrayInitSize();
+        if (maxEntrySize == -1) {
+            maxEntrySize = MAX_ENTRY_LENGTH;
+        }
+        try (ZipInputStream zis = new ZipInputStream(file)) {
+            ZipEntry entry;
+            while ((entry = zis.getNextEntry()) != null) {
```
oanalavinia, line 103: Is this impacting performance?

Issue #592, opened 2025-07-29
Replace deprecated method MathTool.toInteger in Confluence.Macros.Contributors


sarahxb:
https://github.com/xwikisas/xwiki-pro-macros/blob/main/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Contributors.xml#L466

 - Deprecated usage of method [org.apache.velocity.tools.generic.MathTool.toInteger] in xwiki:Confluence.Macros.Contributors@110,117

oanalavinia:
Note that this breaks the functional tests

PR #612, opened 2025-08-21
Replace deprecated method MathTool.toInteger in Confluence.Macros.Contributors #592

sarahxb:
Replace deprecated method MathTool.toInteger in Confluence.Macros.Contributors #592

### DIFF
--- a/xwiki-pro-macros-api/src/main/resources/css/userlist.css
+++ b/xwiki-pro-macros-api/src/main/resources/css/userlist.css
@@ -51,7 +51,6 @@
 
   table.xwiki-userlist tr::after {
     clear: both;
-    content: ;
     display: block;
     content: "";
   }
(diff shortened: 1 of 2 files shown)


xwiki-pro-macros-api/src/main/resources/css/userlist.css
```diff
@@ -51,7 +51,6 @@
 
   table.xwiki-userlist tr::after {
     clear: both;
-    content: ;
```
Farcasut, line 54: I think this should not be part of this PR.

Farcasut (changes requested):

xwiki-pro-macros-api/src/main/resources/css/userlist.css
sarahxb: Yes, I'm not sure how that happened:)

PR #618, opened 2025-09-03
Thumbnails fail to be saved on Windows because of invalid file path #619

ChiuchiuSorin:
Adapted the thumbnail generator to use the hash code computed from the document reference and attachment name to identify the thumbnails.

### DIFF
--- a/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/ThumbnailGenerator.java
+++ b/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/ThumbnailGenerator.java
@@ -71,6 +71,11 @@ import net.coobird.thumbnailator.Thumbnails;
 @Singleton
 public class ThumbnailGenerator
 {
+    /**
+     * Path to the temporary folder where thumbnails are stored.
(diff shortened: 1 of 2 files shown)


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/ThumbnailGenerator.java
```diff
@@ -279,4 +264,34 @@ private boolean isOfficeServerConnected()
...
+            tempDir.mkdirs();
+        } else {
+            // Check if the old directory format exists and delete it. To be removed in a few months.
+            tempDir.listFiles();
+            for (File oldFolder : tempDir.listFiles()) {
+                if (oldFolder.isDirectory()) {
+                    deleteOldFormat(oldFolder);
+                }
+            }
+        }
```
Farcasut, line 283: I think the first tempDir.listFiles() is a leftover

ChiuchiuSorin:
Other then hashing (for which we could use SHA256 to avoid collisions), we could use a URL encoder to ensure that the used symbols are safe for Windows. WDYT?

PR #650, opened 2025-10-24
View File macro - improve performance of thumbnail generator #649

ChiuchiuSorin:
Replaced pdf converter to JPG converter which removes the PDF->JPEG conversion need.

Observed performance changes:

- **Old:**
  - First execution time: **1:14** minutes
  - Second execution time (thumbnails are already existing) : 0:21 
- **New:**
  - First execution time: **44** seconds
  - Second execution time (thumbnails are already existing) : 0:21

### DIFF
--- a/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/thumbnail/generators/OfficeThumbnailGenerator.java
+++ b/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/thumbnail/generators/OfficeThumbnailGenerator.java
@@ -59,7 +59,7 @@ public class OfficeThumbnailGenerator extends AbstractOfficePdfThumbnailGenerato
     public static final String HINT = "office";
 
     @Inject
-    private OfficeManagerWrapper officeFactory;
+    private OfficeThumbnailGeneratorUtils officeUtils;
 
     @Inject
     private Logger logger;
@@ -67,7 +67,7 @@ public class OfficeThumbnailGenerator extends AbstractOfficePdfThumbnailGenerato
     @Override
     public String generateThumbnail(AttachmentReference attachmentReference) throws Exception
     {
-        if (officeFactory.isOfficeServerConnected()) {
+        if (officeUtils.isOfficeServerConnected()) {
             byte[] bais = getJPEGContent(attachmentReference);
             return saveThumbnail(getBufferedImage(bais), attachmentReference);
         } else {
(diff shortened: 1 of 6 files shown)


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/ThumbnailGenerator.java
```diff
@@ -156,45 +156,73 @@ private byte[] getPDFThumbnailBytes(AttachmentReference attachmentReference) thr
...
             manager.stop();
-            return new ByteArrayInputStream(baos.toByteArray());
+            return baos.toByteArray();
         }
     }
+//    first time old: 1:14
+//    second time: 0:21
+//
+//    new: first time: 1:04
+//    second time: 0:21
```
oanalavinia: Leftover

oanalavinia:
Just noticed this was draft, sorry if it wasn't yet ready for reviews :)

oanalavinia (commented):
This looks ok to me. Just note that I haven't looked at every detail. Let me know if you are not sure about some aspects.

What's missing are the tests, unit and functional

ChiuchiuSorin:
I did not write unit tests, because none of the other macros from the `api` module have unit tests implemented.
The view file does have functional tests though: https://github.com/xwikisas/xwiki-pro-macros/blob/main/xwiki-pro-macros-test/xwiki-pro-macros-test-docker/src/test/it/com/xwiki/pro/test/ui/ViewFileIT.java

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/viewfile/internal/thumbnail/generators/OfficeThumbnailGenerator.java
```diff
@@ -61,46 +51,23 @@
...
-     * The office server configuration.
-     */
     @Inject
-    private OfficeServerConfiguration officeServerConfig;
-
-    @Inject
-    private OfficeServer officeServer;
-
-    private OfficeManager officeManager;
+    private OfficeManagerWrapper officeFactory;
```
oanalavinia: Factory is not a very good name, as it gives the idea of having a office factory (like the factory design pattern) and I don't think this is what you have in here. Maybe you want to rename the class altogether as `OfficeUtils` / `OfficeManagerUtils`? I know in general is good to group by a higher purpose, to have specific classes., while Utils is way too generic and not that desired. But in your case you simply want to refactor so you can test as much as possible. So Utils sounds more appropriate in here to me
WDYT?

Issue #644, opened 2025-10-21
You can't sort the details summary macro based on labels, author, doc title or creation date


Farcasut:
**Describe the bug**
You can't sort the details summary macro based on labels, author or creation date 

**Steps to reproduce**
Steps to reproduce the problematic behavior:
1. Add a details summary to a page
2. Sort after the label
3. Save and view

**Expected behavior**

The rows being sorted after the labels field

**Actual behavior**
The rows are not sorted

**Screenshots**

**Additional context**

PR #659, opened 2025-11-03
You can't sort the details summary macro based on labels, author, doc title or creation date #644

Farcasut:
Added the ability to sort on all columns.
Currently, there is an issue when you try to sort a column after the special columns. If the details macro has already defined that column, sorting will occur after the column in the details instead of after the special column. We can define special keywords for the special columns, like @\document,\@\author, @\modified, @\tags, but I'm not sure since this is a bridge and is not meant to be used in xwiki


xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-api/src/main/java/com/xwiki/macros/confluence/detailssummary/internal/macro/ConfluenceSummaryProcessor.java
```diff
@@ -142,9 +141,11 @@ public List<List<Block>> getDetails(String id, List<String> headings, List<Block
         List<List<Block>> rows = new ArrayList<>();
         for (XDOM detailMacro : details) {
             List<Block> row = getRow(detailMacro, headings, columns, columnsLower, doc.getSyntax());
-            if (row != null) {
+            // If the row is empty, it means that either the user provided a list of headers and none were found
```
Farcasut: Not sure if i should keep the row or not

raphj: I would keep it if at least the document title / reference is displayed

Issue #635, opened 2025-10-02
"Page Tree" macro and "Button" macro have their links broken when the page is moved


LucianCheve:
**Steps to reproduce**

1. Have a confluence tree macro and/or a button macro
2. Move the page that is referenced in the confluence tree macro and/or button macro
3. Check the page after the move

**Expected behavior**
All the links in both the tree macro and button macro get updated after the move and still work.

**Actual behavior**
The links are broken and no longer work 

Tested on XWiki 16.10.11 and Pro Macros - Confluence bridges 1.27.3

PR #663, opened 2025-11-18
"Page Tree" macro and "Button" macro have their links broken when the page is moved #635

rbancescu:


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/internal/updateReferences/listeners/PageRenameMacroParametersUpdateListener.java
```diff
@@ -0,0 +1,164 @@
...
+@Named(PageRenameMacroParametersUpdateListener.ROLE_HINT)
+@Singleton
+public class PageRenameMacroParametersUpdateListener extends AbstractEventListener implements Disposable
+{
+    protected static final String ROLE_HINT = "PageRenameMacroParametersUpdateListener";
+
+    private static final Map<String, List<String>> MACROS_TO_UPDATE = Map.of(
+        "button", List.of("url"),
+        "confluence_pagetree", List.of("root")
+        );
```
Farcasut: Pleasea add a comment

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/internal/updateReferences/ReferencesSolrMetadataExtractor.java
```diff
@@ -0,0 +1,202 @@
...
+    @Inject
+    private Logger logger;
+
+    @Override
+    public boolean extract(XWikiDocument document, SolrInputDocument solrDocument)
+    {
+        XDOM xdom = document.getXDOM();
+        boolean linksRegistered = false;
+
+        for (Map.Entry<String, List<String>> macroToSearch : MACROS_TO_SEARCH.entrySet()) {
```
raphj: I think this misses macros that are in other macros. If I'm not wrong, you have to recursively parse the content of all the macros and add your specific handling for the macros you want to handle. 

See for instance https://snippets.xwiki.org/xwiki/bin/view/Extension/RemoveMacrosFromXDOM/ for an example of code that searches a document content for macros.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/internal/updateReferences/listeners/PageRenameMacroParametersUpdateListener.java
```diff
@@ -0,0 +1,164 @@
...
+                }
+            }
+        }
+    }
+
+    @Override
+    public void dispose()
+    {
+        try {
+            stopThread(this.referencesMacroThread, this.referencesMacroRunnable);
```
raphj: Same issue in the diagram application, but it seems to me we are starting a thread that will be sitting there forever (until the component is updated) as soon as a document is renamed.

Isn't this an issue? Shouldn't we make the thread leave when it doesn't have anything to do, maybe after some time?

Naively, I would think we could drop this heavy thread handling and use something like [CompletableFuture.runAsync](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#runAsync-java.lang.Runnable-)

cc @oanalavinia

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/internal/updateReferences/ReferencesSolrMetadataExtractor.java
Farcasut: I think you should use an OrBlockMatcher to search for all the macros at once, then check their names against the map and update only the necessary parameters.

Farcasut (changes requested):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/internal/updateReferences/ReferencesSolrMetadataExtractor.java
michitux: > I think using `Block.Axes.DESCENDANT` as a parameter solves the problem!

No, it doesn't. Implementing `MacroRefactoring` would solve this as then this whole complexity is handled by XWiki and you don't need to re-implement it.

michitux:
> @michitux I have the implementation [here](xwiki-pro-macros-api/src/main/java/com/xwiki/macros/internal/updateReferences/UpdateParametersRefactoring.java), which is used [here ](https://github.com/xwikisas/xwiki-pro-macros/pull/663/files#diff-6056fd5e8f0617e3a3d5722111965a1ffa65b14d30bbb80245b0d06bf373029eR117). I should have approached this in a different way?

Yes, you should have included the code for extracting the references in that component by implementing `extractReferences`. That would completely replace all the code for Solr indexing as it would automatically be called during the extraction of backlinks. Further, the component should be named (I mean the `@Named` annotation) exactly like the macro, then it will automatically be called when a page is moved (and during indexing) so there shouldn't be any need to call it manually unless you need to provide a way to fix existing pages outside a move operation.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/internal/updateReferences/AbstractReferenceUpdateMacroRefactoring.java
```diff
@@ -0,0 +1,146 @@
...
+                        sourceReference);
+
+                boolean resolvedRelative = !isReferenceAbsolute(stringMacroReference, macroReference);
+
+                if (macroReference.equals(sourceReference)) {
+                    newMacroBlock.setParameter(parameterToUpdate,
+                        serializeTargetReference(targetReference, currentDocumentReference, resolvedRelative));
+                }
+            }
+            return Optional.of(newMacroBlock);
```
michitux: I'm not sure but I think it would be better to only return a non-empty result when the macro block has actually been changed.

xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-api/src/main/java/com/xwiki/macros/confluence/confluencepagetree/internal/macro/ConfluencePageTreeMacroParametersRefactoring.java
```diff
@@ -0,0 +1,150 @@
...
+    {
+        String rootValue = macroBlock.getParameter(ROOT_PARAMETER);
+
+        XWikiContext xContext = contextProvider.get();
+        XWiki xWiki = xContext.getWiki();
+
+        try {
+            DocumentReference docRef = documentReferenceResolver.resolve(rootValue);
+
+            if (xWiki.exists(docRef, xContext)) {
```
michitux: I don't think it's correct to check if the page exists here, this code should just return all possible references. What if the target page is created after this macro has been indexed?

Issue #647, opened 2025-10-24
TOC entries links don't work with Safari when expand is used and sections are not expanded


iuliabalan:
**Describe the bug**
When using Safari and clicking on any of the TOC entries nothing happens if the expand macro is used and sections are not expanded.

**Steps to reproduce**
Add some content with headings to a page then add a TOC. For example:

```
{{toc/}}

= Section1 =

Tests

== Section 1.1 ==

Tests

= Section 2 =

Tests

{{expand title="Section 3"}}

= Section 3 =

More tests

Lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec auctor urna accumsan, fringilla massa sit amet, suscipit arcu. Pellentesque a sem fringilla, luctus eros sed, ullamcorper neque. Nunc molestie mollis convallis. Duis quis arcu sem. Fusce tellus turpis, ultricies vel purus sed, cursus hendrerit felis. Aenean malesuada convallis leo. Donec blandit tempus elit eu pellentesque. Sed suscipit a dui eu dictum. Praesent sagittis malesuada sollicitudin. Maecenas maximus nisl augue, in gravida nisi luctus lobortis. Mauris quis convallis dui, eget eleifend orci. Sed at auctor magna. Vivamus efficitur venenatis posuere.

{{/expand}}
```

Observe the results with Chrome version 141:
* clicking on any of the entries of TOC redirects the use to that particular section from the page
* also when hovering on any of the TOC entries the link to the section is displayed on the bottom left

Observe the result with Safari (latest version 18.6)
*  clicking on any of the TOC entries that don't have expand work well (e.g go to TOC and click on Section1, Section2 and you'll be redirected to them)
* clicking on TOC entries that are inside expand macro doesn't work unless you go first to that section entry and expand it (e.g click on Section3 => nothing happens, then go to Section 3 of the page, expand it, go to TOC and access it from there=>it works)
* hovering on the TOC entry doesn’t show the link to the corresponding section on the bottom left of the page

I've added more dummy content for visual effect, see my results with Safari:

https://github.com/user-attachments/assets/5eb483f0-be8f-458c-baa7-c41cbeb8c2a8

Farcasut:
Safari 26.2 has been released for nearly three months now. I don't know if I should remove the workaround since it's been enough time for everyone to update.

snazare:
[...]

PR #674, opened 2025-12-09
TOC entries links don't work with Safari when expand is used and sections are not expanded #647

Farcasut:
The issue is that Apple's Safari doesn't respect the HTML standard and doesn't automatically expand `<details>` elements when a URL fragment targets an element inside them. This issue has been fixed in the WebKit engine that Safari uses, but it hasn't reached Safari yet, so other browsers that use WebKit should work correctly. To fix the issue, I added a small js snippet that checks whether the current browser is Safari, and if it is, we perform the expansion ourselves.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
@@ -145,10 +145,7 @@ Hello
 // and we should follow it so we know when the fix reaches Safari. I also opened
 // https://github.com/xwikisas/xwiki-pro-macros/issues/682 on our side to keep track of the issue and avoid forgetting it.
 require(['jquery'], function ($) {
-	var isSafari = navigator.vendor &amp;&amp; navigator.vendor.indexOf('Apple') &gt; -1 &amp;&amp;
-		navigator.userAgent &amp;&amp;
-		navigator.userAgent.indexOf('CriOS') == -1 &amp;&amp;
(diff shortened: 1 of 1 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/Expand.xml
```diff
@@ -51,6 +51,137 @@ Note: if no title is provided, the title of the first panel or the first header
...
+      <code>// The current versions of Safari don't bundle the latest version of WebKit, and because of this it lacks
+// a feature from the HTML standard where the details tag automatically expands when the URL fragment
+// targets an element inside it. Here is the WebKit bug report: https://bugs.webkit.org/show_bug.cgi?id=228843,
+// and we should follow it so we know when the fix reaches Safari. I also opened
+// https://github.com/xwikisas/xwiki-pro-macros/issues/682 on our side to keep track of the issue and avoid forgetting it.
+require(['jquery'], function ($) {
+	var isSafari = navigator.vendor &amp;&amp; navigator.vendor.indexOf('Apple') &gt; -1 &amp;&amp;
+		navigator.userAgent &amp;&amp;
+		navigator.userAgent.indexOf('CriOS') == -1 &amp;&amp;
+		navigator.userAgent.indexOf('FxiOS') == -1;
```
raphj: I don't think we should filter out Chrome and Firefox on iOS as they are supposed to use the exact same webkit version as Safari iOS

Note that you can use .includes() instead of .indexOf(...) !== -1 or  .indexOf(...) > -1

PR #708, opened 2026-02-10
Refactor some code to avoid suppressing checkstyle rules

raphj:
I have refactored some code to remove checkstyle suppressions.


Josue-T (approved):
Didn't check really deeply but at first look it like good.

raphj:
Thanks!

PR #717, opened 2026-03-09
Named colors are not taken in account in the button macro #683

### LINKED ISSUE #683: Named colors are not taken in acount in the button macro
**Describe the bug**

`{{button color="blue"/}}` doesn't make the button blue

**Steps to reproduce**
1. Add the button macro with the wysiwyg editor
2. Type blue for the color
3. Notice that the blue color is indeed picked up by the parameter picker

**Expected behavior**

The blue parameter results in a blue button

**Actual behavior**

The button remains uncolored

sarahxb:
fixes #683

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
@@ -380,6 +380,7 @@ $services.localization.render('rendering.macro.button.description')
         ## The factor value was found empirically.
           filter: brightness(0.85);
           color: white;
+          ## See https://css-tricks.com/approximating-contrast-color-with-other-css-features/
           color: oklch(from $bgColor round(1.21 - L) 0 0);
         #end
       #end
--- a/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
@@ -325,9 +325,6 @@ $services.localization.render('rendering.macro.button.description')
   #set ($title = $escapetool.xml($wikimacro.parameters.title))
   #set ($url = $escapetool.xml($wikimacro.parameters.url))
   #set ($color = $wikimacro.parameters.color)
-  #if(!$wikimacro.parameters.color &amp;&amp; $!xcontext.macro.params != "")
-    #set($color = $xcontext.macro.params)
-  #end
   #set ($width = $escapetool.xml($wikimacro.parameters.width))
   #set ($newTab = $escapetool.xml($wikimacro.parameters.newTab))
   #set ($icon = $escapetool.xml($xcontext.macro.params.icon))
@@ -400,8 +397,6 @@ $services.localization.render('rendering.macro.button.description')
 ## don't have view right on those pages.
 #if ($services.promacrolicensing.hasLicensureForEntity($xcontext.macro.doc.documentReference))
   #executeMacro
-  $wikimacro.parameters.color
-  $xcontext.macro.params.color
 #else
   {{missingLicenseMessage extensionName="proMacros.extension.name"/}}
 #end
--- a/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
@@ -311,7 +311,6 @@ $services.localization.render('rendering.macro.button.description')
   #set ($title = $escapetool.xml($wikimacro.parameters.title))
   #set ($url = $escapetool.xml($wikimacro.parameters.url))
   #set ($color = $wikimacro.parameters.color)
-  ## CSS string color version
   #set ($colorCss = $wikimacro.parameters.get("color"))
   #if (!$colorCss)
     #set ($colorCss = $xcontext.macro.params.get("color"))
@@ -338,6 +337,7 @@ $services.localization.render('rendering.macro.button.description')
     #darken($colors, $darkenedColors)
     #set ($bgColor = "rgb($colors[0],$colors[1],$colors[2])")
     #set ($borderColor = "rgb($darkenedColors[0],$darkenedColors[1],$darkenedColors[2])")
+    ## Compute perceived brightness using standard RGB weighting (WCAG formula).
     #set ($intensity = $colors[0]*0.299 + $colors[1]*0.587 + $colors[2]*0.114)
     #if ($intensity &gt; 186)
       #set ($textColor = "#333")
@@ -377,6 +377,7 @@ $services.localization.render('rendering.macro.button.description')
         border-color: $borderColor;
         color: $textColor;
         #if (!$isColorObject)
+        ## The factor value was found empirically.
           filter: brightness(0.85);
         #end
       #end
--- a/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
@@ -379,6 +379,8 @@ $services.localization.render('rendering.macro.button.description')
         #if (!$isColorObject)
         ## The factor value was found empirically.
           filter: brightness(0.85);
+          color: white;
+          color: oklch(from $bgColor round(1.21 - L) 0 0);
         #end
       #end
       width: $!width;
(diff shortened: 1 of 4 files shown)


xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
```diff
@@ -337,10 +328,26 @@ $services.localization.render('rendering.macro.button.description')
...
+    #if ($intensity &gt; 186)
+      #set ($textColor = "#333")
+    #else
+      #set ($textColor = "#fff")
+    #end
   #else
-    #set ($colors = [255, 255, 255])
+    #set ($bgColor = $colorCss)
+    #set ($borderColor = $colorCss)
+    #set ($textColor = "white")
```
raphj, line 350: Seems like it should be  a function of colorCss.

See for instance the following discussions about this (I have not checked them thoroughly, they come up with a "css pick between black and white text background" web search). It should be possible to find a CSS expression that extracts the relevant data and choses the right color, with possibly a fallback to white.
- https://til.jakelazaroff.com/css/swap-between-black-and-white-text-based-on-background-color/
- https://stackoverflow.com/questions/3942878/how-to-decide-font-color-in-white-or-black-depending-on-background-color

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
```diff
@@ -337,10 +328,26 @@ $services.localization.render('rendering.macro.button.description')
...
+  #if ($color &amp;&amp; $color.class.name == "java.awt.Color")
+    #set ($isColorObject = true)
+  #end
+  #if ($isColorObject)
     #set ($colors = [$color.Red, $color.Green, $color.Blue])
+    #set ($darkenedColors = [])
+    #darken($colors, $darkenedColors)
+    #set ($bgColor = "rgb($colors[0],$colors[1],$colors[2])")
+    #set ($borderColor = "rgb($darkenedColors[0],$darkenedColors[1],$darkenedColors[2])")
+    #set ($intensity = $colors[0]*0.299 + $colors[1]*0.587 + $colors[2]*0.114)
```
raphj, line 341: I would request that the magic constant be explained  (I notice they are the same as the ones mentioned at https://stackoverflow.com/questions/3942878/how-to-decide-font-color-in-white-or-black-depending-on-background-color), but I understand that this comes from code you moved and this was already there. I'll let you decide if you want to clarify or leave this like this and consider this out of scope of your contribution.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
```diff
@@ -305,26 +305,17 @@ $services.localization.render('rendering.macro.button.description')
...
-    #fff
-  #end
-#end
 #macro (executeMacro)
   ## Retrieve macro parameters
   #set ($label = $escapetool.xml($wikimacro.parameters.label))
   #set ($title = $escapetool.xml($wikimacro.parameters.title))
   #set ($url = $escapetool.xml($wikimacro.parameters.url))
   #set ($color = $wikimacro.parameters.color)
+  ## CSS string color version
```
raphj: That comment looks a bit unclear to me, I think it can be dropped altogether (the code looks clear enough without it).

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
```diff
@@ -367,13 +372,16 @@ $services.localization.render('rendering.macro.button.description')
...
-        #if ("$!color" != '')
-        background-color: #toCssRGB($colors);
-        border-color: #toCssRGB($darkenedColors);
-        color: #fontColor($colors);
+      #if ("$!bgColor" != '')
+        background-color: $bgColor;
+        border-color: $borderColor;
+        color: $textColor;
+        #if (!$isColorObject)
+          filter: brightness(0.85);
```
raphj, line 381: Where does the 0.85 factor come from?

raphj (changes requested):
Have you done some accessibility checks?

Colors are easy to get wrong. Especially when involving text AND background color.

I'm also well aware that correctly handling every possible arbitrary color coming from users is likely an impossible goal and we may consider that users have _some_ responsibility in picking colors that work.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
sarahxb: It's a bit complicated here, because the mathematical solutions require to have the color in RGB, there is this function https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/contrast-color , but it's not very stable

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
sarahxb: I chose  0.85 because it approximates the visual effect of the previous darken(r,g,b) logic. It makes it 15% darker. it was mostly trial and error here

raphj: Okay, maybe add a comment explaining that this value was found empirically (which is fine btw)

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml
raphj: Perhaps this? https://css-tricks.com/approximating-contrast-color-with-other-css-features/

sarahxb: It's still a bit risky if it's an older browser https://css-tricks.com/almanac/functions/o/oklch/  , it normally uses an RGB fallback, but I guess it will be just white in this case. But I can give it a try

sarahxb: > Perhaps this? https://css-tricks.com/approximating-contrast-color-with-other-css-features/

image

this is the result :(

```
 #if (!$isColorObject)
          filter: brightness(0.85);
          color: oklch(from $bgColor calc((0.21 - l) * 100%) 0 0);
        #end
  ```

Comment (raphj) [reply xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml:350]: @sarahxb And with `oklch(from $bgColor round(1.21 - L) 0 0)`? With a color picker I found you used `#0000d9` as a background, your expression indeed gives me black but mine gives me white (Firefox):

image

Comment (raphj) [reply xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml:350]: For handling older browsers, I suppose you'd need to set color twice, with the fallback first, like this:

```
 #if (!$isColorObject)
          filter: brightness(0.85);
          color: white;
          color: oklch(from $bgColor calc((0.21 - l) * 100%) 0 0);
        #end
```

You could even have it 3 times, with the last one using `contrast-color`.

Comment (sarahxb) [reply xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml:350]: image

have you also tried with lighter colors?

Comment (raphj) [reply xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml:350]: Whoops, I copy pasted the wrong expression. With `oklch(from $bgColor round(1.21 - L) 0 0)` I get nice results for each of these colors! @sarahxb

**Review by raphj — APPROVED** [review]

**Review by raphj — APPROVED** [review]
I suppose It's hard to actually test the result of the oklch expression, right?

**Comment by sarahxb** [comment]
> result of the oklch expression

@raphj With a comment explaining the expression is this ok ```assertEquals("oklch(1 0 0)",btn6.getFontColor()); 
 assertEquals("oklch(0 0 0)",btn7.getFontColor());``` ?

**Comment by raphj** [comment]
> > result of the oklch expression
> 
> @raphj With a comment explaining the expression is this ok `assertEquals("oklch(1 0 0)",btn6.getFontColor()); assertEquals("oklch(0 0 0)",btn7.getFontColor());` ?

@sarahxb Ah, nice, you can get the computed value of the L axis! Actually seems like a great idea, if it's not too much work you could test a few colors like this. 2 is already great!
A comment is a good idea, the both of us would be very familiar with those asserts but they could easily look a bit obscure (ah!) to others.

File: xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Button.xml [hunk 5]
```diff
@@ -367,13 +372,19 @@ $services.localization.render('rendering.macro.button.description')
...
-        color: #fontColor($colors);
+      #if ("$!bgColor" != '')
+        background-color: $bgColor;
+        border-color: $borderColor;
+        color: $textColor;
+        #if (!$isColorObject)
+        ## The factor value was found empirically.
+          filter: brightness(0.85);
+          color: white;
+          color: oklch(from $bgColor round(1.21 - L) 0 0);
```
raphj, line 384:
```suggestion
          ## See https://css-tricks.com/approximating-contrast-color-with-other-css-features/
          color: oklch(from $bgColor round(1.21 - L) 0 0);
```

raphj (approved):

Issue #732, opened 2026-04-17
Improve Recently Updated parameter handling


raphj:
The Recently Updated macro uses complex json parameter handling that can be simplified a bit. The structure can be flattened and then the parameters can be passed using standard HTTP post parameters instead of going through JSON, which improves performance and clarity a bit.

PR #733, opened 2026-04-17
#732: Improve Recently Updated parameter handling

raphj:
See #732. Some simplification to the recently updated parameter handling.

### DIFF
--- a/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdated.xml
+++ b/xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdated.xml
@@ -173,7 +173,11 @@ There are only a few things that are not supported compared to the Confluence ma
 require(['jquery'], $ =&gt; {
   const MACRO_SERVICE_URL = new XWiki.Document(
     XWiki.Model.resolve('Confluence.Macros.RecentlyUpdatedService', XWiki.EntityType.DOCUMENT)
-  ).getURL('get');
+  ).getURL('get', $.param({
+    xpage: 'plain',
+    outputSyntax: 'plain',
+    fromAjax: true
+  })) + '&amp;';
 
   /*
    * Query new results and append them to the page
@@ -185,14 +189,7 @@ require(['jquery'], $ =&gt; {
       return;
     }
 
-    const urlParams = [
-      'xpage=plain',
-      'outputSyntax=plain',
-      'fromAjax=true',
-      'macroOptions=' + encodeURIComponent(macro.dataset.options),
-    ].join('&amp;');
-    const url = MACRO_SERVICE_URL + '?' + urlParams;
-    fetch(url)
+    fetch(MACRO_SERVICE_URL + macro.dataset.options)
     .then(function (response) {
       if (!response.ok) {
         throw new Error('An error occured fetching new results.');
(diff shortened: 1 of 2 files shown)


xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdated.xml
```diff
@@ -185,13 +185,7 @@ require(['jquery'], $ =&gt; {
...
     }
 
-    const urlParams = [
-      'xpage=plain',
-      'outputSyntax=plain',
-      'fromAjax=true',
-      'macroOptions=' + encodeURIComponent(macro.dataset.options),
-    ].join('&amp;');
-    const url = MACRO_SERVICE_URL + '?' + urlParams;
+    const url = MACRO_SERVICE_URL + '?xpage=plain&amp;outputSyntax=plain&amp;fromAjax=true&amp;' + macro.dataset.options;
```
Farcasut: I feel that we could use this method of creating the query string to make it easier to read, but feel free to ignore this suggestion since it's just a personal preference.
```
    const QUERY_STRING = $.param({
        'xpage': plain,
        'outputSyntax': plain,
        'fromAjax': true,
        'macroOptions': macro.dataset.options
      });
   const url = `${MACRO_SERVICE_URL}?${QUERY_STRING}`
```

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdatedService.xml
```diff
@@ -451,17 +450,37 @@ $stringtool.replaceEach($v, $solrSpecialChars, $escapedSolrSpecialChars)##
...
+  })
+  ## Boolean parameters
+  #foreach ($k in ["showProfilePic", "page-hasNext", "blogpost-hasNext", "attachment-hasNext", "comment-hasNext"])
+    #if ("$!request[$k]" != "")
+      #set ($options[$k] = $request[$k] == "true")
+    #end
+  #end
   #fetchResults($options, $results)
-  #displayResults($results $options)
+  #displayResults($results, $options)
```
Farcasut, line 483: You should use ' if the string doesn't require interpretation

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdated.xml
raphj: Done! This change was actually done because "macroOptions" is not a parameter anymore. I changed the code to move that fix part to the constant and got rid of the QUERY_STRING constant altogether.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdatedService.xml
raphj: Done. I addressed this for the whole code while I was at it.

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdatedService.xml
```diff
@@ -69,22 +69,22 @@ $stringtool.replaceEach($v, $solrSpecialChars, $escapedSolrSpecialChars)##
 
 #macro (addSolrParameterListWithSign $listString $fq $fieldName $onlyComma, $forAuthors)
   #if ("$!listString" != '')
-    #set ($solrList = " ")
-    #set ($sep = "[\s]*,[\s]*|[\s]+")
-    #if ("$onlyComma" == "true")
-      #set ($sep = ",")
+    #set ($solrList = ' ')
+    #set ($sep = '[\s]*,[\s]*|[\s]+')
+    #if ('$onlyComma' == 'true')
```
Farcasut: Here you have to use "" to interpret the $onlyComma as a string because right now you are comparing $onlyComma == true. You might be able to use booleans to do the compassion, but to be honest I don't remember.

Farcasut (changes requested):

xwiki-pro-macros-ui/src/main/resources/Confluence/Macros/RecentlyUpdatedService.xml
raphj: I fixed this, but I'd prefer not to mess too much with the existing code here to be honest, it's a bit out of scope of this PR

edit: actually, good catch, I would have broken this!

Farcasut (approved):

PR #735, opened 2026-04-20
#707: Upgrade XWiki Parent to 15.10

raphj:
- [ ] Note: please do not merge before #733 is merged and a new version of Pro Macros is released with it.


Farcasut (commented):
You should also export all the pages to have the correct xml version and new properties

raphj:
> You should also export all the pages to have the correct xml version and new properties

Done!

PR #762, opened 2026-04-30
Cannot type after adding the status macro in the beginning of a line / paragraph #725

### LINKED ISSUE #725: Cannot type after adding the status macro in the beginning of a line / paragraph
**Steps to reproduce**
1. create a page or edit an existing page
2. In the rich text editor, go at the beginning of a paragraph (new line), and add a status macro (with the inline / command or through the menu)
3. [optional] configure the macro parameters (color, text)

**Expected behaviour**
I would like to add some more text after my status macro, to type next to it on the right

**Actual behavior**
It is not possible to position the caret on the right of the status macro to type some more text after, the status macro is taking the whole line

**Screenshots**

Image

**Environment (please complete the following information):**
 - OS: Ubuntu
 - Browser : Chrome, but probably reproduces on the other browsers in the same way
 - XWiki version 17.10.2
 - App version 1.29.0

**Additional context**
There was a very similar issue with the mentions on the XWiki product, caused by the same underlying editor limitation, probably: https://jira.xwiki.org/browse/XWIKI-17965 .

ChiuchiuSorin:
Changes done:
* Moved the status macro from XObject to java.
* Forced inline mode by replacing the macro when used in edit mode and not inline.

UI changes:
* The `subtle` parameter is now displayed as a checkbox instead of text input:

* Old view:
old

* New view:
new


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
```diff
@@ -0,0 +1,143 @@
...
+            StringBuilder cssClass = prepareCSS(parameters);
+
+            Map<String, String> blockParameters = Map.of("class", cssClass.toString());
+            List<Block> blocks = this.contentParser.parse(getTitle(parameters), context, false, true).getChildren();
+            blocks = List.of(new FormatBlock(blocks, Format.NONE, blockParameters));
+
+            return blocks;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
```
raphj, line 111: Why catch Exception to throw RuntimeException? Wouldn't it be better not to try-catch at all here?

If you decide to keep this, maybe add a comment explaining why it's there.

raphj (approved):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
```diff
@@ -0,0 +1,143 @@
...
+                Block mockParent = new CompositeBlock();
+                mockParent.addChild(oldMacroBlock);
+
+                return List.of();
+            }
+
+            StringBuilder cssClass = prepareCSS(parameters);
+
+            Map<String, String> blockParameters = Map.of("class", cssClass.toString());
+            List<Block> blocks = this.contentParser.parse(getTitle(parameters), context, false, true).getChildren();
```
michitux: Should the title really be treated as wiki syntax? Or shouldn't this rather be parsed with a plain text parser?

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
ChiuchiuSorin: Indeed, I forgot to log the error. I added it now.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
ChiuchiuSorin: I modified the content parser with a word block to avoid treating it as wiki syntax.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
raphj: Throwing RuntimeException looks a bit weird to me. I don't really understand why we can't let the code calling internalExecute deal with the unexpected exceptions and the error reports. 

I see now that view-file also does this, I don't think it is a really good idea and most macros don't do this.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
```diff
@@ -0,0 +1,133 @@
...
+                Block mockParent = new CompositeBlock();
+                mockParent.addChild(oldMacroBlock);
+
+                return List.of();
+            }
+
+            StringBuilder cssClass = prepareCSS(parameters);
+
+            Map<String, String> blockParameters = Map.of("class", cssClass.toString());
+            WordBlock wordBlock = new WordBlock(getTitle(parameters));
```
raphj: Using WordBlock like this is not right. For instance a word block is not supposed to contain spaces or special characters.

You should use a plain text parser. See for instance how this is done at https://github.com/xwikisas/xwiki-pro-macros/blob/489162ef7d58b25ab93ee738ea855abd9fadf2ba/xwiki-pro-macros-api/src/main/java/com/xwiki/macros/tab/internal/TabMacro.java#L172

raphj (changes requested):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
```diff
@@ -0,0 +1,129 @@
...
+        List<Block> blocks =
+            this.contentParser.parse(getTitle(parameters), Syntax.PLAIN_1_0, context, false, null, true).getChildren();
+        blocks = List.of(new FormatBlock(blocks, Format.NONE, blockParameters));
+        return blocks;
+    }
+
+    private String getTitle(StatusMacroParameters parameters)
+    {
+        String title = parameters.getTitle();
+        if (title == null || title.isEmpty()) {
```
michitux:
```suggestion
        if (StringUtils.isBlank(title)) {
```

Your version is correct, it just seems simpler with `StringUtils` and I think it is nice to also not use blank (whitespace) titles. Alternatively, there is of course also `StringUtils#isEmpty` if you really only want to catch `null` and empty strings.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
ChiuchiuSorin: I tried using the `PlainTextBlockParser` as suggested, but it returns a paragraph block which breaks the UI. An easy workaround that I found is to use the `MacroContentParser` with the `Syntax.PLAIN_1_0`.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    protected List<Block> internalExecute(StatusMacroParameters parameters, String content,
+        MacroTransformationContext context) throws MacroExecutionException
+    {
+        if (!context.isInline() && isEditMode(context)) {
+            // Construct a new macro block to replace the old one with the inline flag set to true.
+            MacroBlock oldMacroBlock = context.getCurrentMacroBlock();
+            MacroBlock replacement =
+                new MacroBlock(oldMacroBlock.getId(), oldMacroBlock.getParameters(), oldMacroBlock.getContent(), true);
+            oldMacroBlock.getParent().replaceChild(new ParagraphBlock(List.of(replacement)), oldMacroBlock);
```
michitux, line 104: I would suggest using a `MacroMarkerBlock` and directly execute the macro, i.e., also compute the children as in the code below and add as children to the `MacroMarkerBlock`. And then also implement `boolean isExecutionIsolated(StatusMacroParameters parameters, String content)` to return `true` (it doesn't matter if the parent version has it or not, you can still implement it without `@Override`). Otherwise, the macro transformation needs to re-scan the whole XDOM to find the new macro block to execute and then call this method again. I know that there are XWiki instances with many (1k or more iirc) status macros on a single page and not being able to safely mark the status macro as isolated would severely impact the rendering performance.

michitux (commented):
Looks good in general, but I have one small code style suggestion and a bigger suggestion regarding performance.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/status/internal/StatusMacro.java
michitux: We have [`org.xwiki.rendering.util.ParserUtils#removeTopLevelParagraph`](https://github.com/xwiki/xwiki-rendering/blob/53e8c57e70ee4f7adaf7f89fccbf0d727a27c344/xwiki-rendering-api/src/main/java/org/xwiki/rendering/util/ParserUtils.java#L46) as a way to deal with this paragraph but what you did is good, too.

michitux (approved):
Looks good to me. Does it work for you as intended?

raphj (approved):

PR #769, opened 2026-05-20
Table not correctly rendered when used into layout macros #767

Farcasut:
The issue seems to happen because of a weird interaction between flex containers, table-layout: auto, and break-word. Flex shrinks by default the items and because of table-layout: auto, the browser then sizes columns proportionally to their content, so the long column takes most of the width, leaving almost nothing for the shorter one and because break-word allows the splitting at each char the column ends up with a single vertical line.
Preview with the fix:
The example from the issue
image

An example with really long single 'words':
image

Some old test for my sanity:

image

image

image

image


raphj (approved):
Looks quite good, your manual tests look quite convincing. It would be nice to find a way to share the test pages somehow (I have no idea how) so someone working on this stuff can check with them.

PR #795, opened 2026-07-09
[#734] Make the Expand macro title parameter mandatory

### LINKED ISSUE #734: UI/UX - Expand Macro - Title is marked as optional even though it's required
Despite being marked as an optional parameter, if left empty the title is automatically set to "Click here to expand...", meaning that the title isn't really an optional parameter.

**Proposed fixes:**
Make the title a required parameter, instead of keeping it optional and working around the limitations this brings.

bLaCkwEw:
## Problem

The `title` parameter of the Expand macro was declared as **optional** (`mandatory = 0`) but had a
default value of `Click here to expand...`. As a result, even though the parameter was presented as
optional, leaving it empty always produced the default title, so it was never really optional.

This is inconsistent and misleading: the UI marks the parameter as optional while the rendering
treats it as required by falling back to a default value.

## Solution

As proposed in the issue, make the `title` parameter a **required** parameter:

- Set `mandatory` to `1` for the `title` parameter of the Expand macro.
- Remove the misleading default value (`Click here to expand...`) from the parameter definition.
- Update the macro documentation:
  - Mark `title` as **Required** in the parameters table.
  - Remove the now-obsolete note describing the auto-title / default-title fallback behavior.
  - Update the example usage to include a `title`.
  - Fix a couple of typos in the macro description (`writting` → `writing`, `collaspse` → `collapse`).
- Clarify the parameter description in the translation files (EN and FR) to mention that the
  parameter is mandatory.


raphj:
@bLaCkwEw @Farcasut what's the status on this?

bLaCkwEw:
Closing PR as @Farcasut took over the issue.

PR #801, opened 2026-07-16
#789: Titles from layout section macros have less space than standard titles

### LINKED ISSUE #789: Titles from layout section macros have less space than standard titles
The titles from layout section macros have a smaller margin / whitespace than standard titles in XWiki pages. When importing content from Confluence, it ends up looking like it's too compressed and there is little to no space between the title and the content:

Image

In the screenshot we can see that there is little space between the title and the content. Manually inserting a white row (to have 2 in total) makes it so the 2 are better separated. Maybe it's too much space, but as it is now it seems there is too little. 

Regardless of whether titles are in a layout section or not, the space that separates them from the next paragraph should be the same

Josue-T:
Fix #789  and also cleanup the duplicate CSS rule.

### DIFF
--- a/xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceLayoutSection.xml
+++ b/xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceLayoutSection.xml
@@ -233,8 +233,7 @@ This is a bridge for the Confluence Layout Section macro. It is usually used wit
 .macro-layout-section {
   gap: 32px;
   h1, h2, h3, h4, h5, h6 {
-    margin: 0;
-    margin-bottom: 10px;
+    margin-top: 0;
   }
 }
 </code>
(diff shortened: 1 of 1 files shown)


xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceLayoutSection.xml
```diff
@@ -234,13 +234,7 @@ This is a bridge for the Confluence Layout Section macro. It is usually used wit
...
   h1, h2, h3, h4, h5, h6 {
     margin: 0;
-  }
-}
-
-.macro-layout-section {
-  gap: 32px;
-  h1, h2, h3, h4, h5, h6 {
-    margin: 0;
+    margin-bottom: 10px;
```
raphj: Could we prehaps have a character-size dependent margin (in `em` maybe?)

raphj (changes requested):

xwiki-pro-macros-confluence-bridges/xwiki-pro-macros-confluence-bridges-ui/src/main/resources/Confluence/Macros/ConfluenceLayoutSection.xml
Josue-T: yeah it's true. Well the idea was mainly to just reuse the default value from XWiki platform. But I think we could also just avoid to override ` margin-bottom` value and only change the ` margin-top` value. I pushed fe4313a for this.

raphj: Using the default XWiki platform value is the right thing to do indeed.

But not overriding it in the first place is even better indeed!

raphj (approved):

Farcasut (approved):

PR #802, opened 2026-07-16
UI/UX - User List Macro - Unclear what properties are available #740

### LINKED ISSUE #740: UI/UX - User List Macro - Unclear what properties are available
Properties are case-sensitive, and sometimes have different names than the ones immediately visible in the UI. For example, to display the "**About**" section of a user profile in a User List table, you have to add the "**comment**" property to the properties list.

**Proposed fixes:**
Fetch the available properties and present them to the user, similar to what the "**Groups**" input does.

Farcasut:
* Added a picker for the properties macro
* Added a config where the admins can ban certain properties with a few banned by default
* Updated the macro parameter to use string instead of the array because there is a bug if you use a multi select picker where the macro modal will not load.

https://github.com/user-attachments/assets/1531c70d-fbcf-4a7e-a3a9-b2b4b77b5bcc


xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+    @Named("userlist")
+    private ConfigurationSource configurationSource;
+
+    @Inject
+    private ContextualAuthorizationManager authorizationManager;
+
+    /**
+     * @return set of banned fields, or an empty set if the current user has admin rights.
+     */
+    public Set<String> bannedFields()
```
raphj: This method looks like a getter and should probably be named #getBannedFields()#.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/internal/macro/UserListConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+    {
+        if (this.authorizationManager.hasAccess(Right.ADMIN)) {
+            return Collections.emptySet();
+        }
+
+        String stringField = this.configurationSource.getProperty("bannedFields", "");
+        return new HashSet<>(List.of(stringField.split(",")));
```
raphj: This probably doesn't matter much (but could still cause surprises later), but it will return a Set containing one element, the empty string, if the configuration is empty, instead of an empty set.

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
```diff
@@ -19,28 +19,28 @@
...
 
 import com.xwiki.macros.internal.grouplist.GroupReferenceList;
 import com.xwiki.macros.internal.userlist.UserReferenceList;
+import com.xwiki.pickers.userList.UserListProperties;
 
 /**
  * Parameters for the {@link com.xwiki.macros.userlist.internal.macro.UserListMacro}.
  *
  * @version $Id$
+ * @since 1.31.4
```
raphj:
```suggestion
 * @since 1.4.0
```
This class was introduced in commit 0a47cf262915a9a18444929fa08bd155a0662fe0, released in Pro Macros 1.4

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
```diff
@@ -106,7 +106,8 @@ public List<String> getProperties()
      */
     @PropertyName("Properties")
     @PropertyDescription("List of user properties to be displayed")
-    public void setProperties(List<String> properties)
+    @PropertyDisplayType(UserListProperties.class)
+    public void setProperties(String properties)
```
raphj, line 110: This is a backward compatibility breakage. Potential external callers (I know, unlikely, but still) of `UserListMacroParameters` expect a `List<String>` for both the setter and the getter. Is there a way to keep this stuff as `List<String>`?

Or maybe it was always broken like this?

xwiki-pro-macros-api/src/main/java/com/xwiki/pickers/userList/UserListProperties.java
```diff
@@ -17,42 +17,15 @@
...
-
-import org.xwiki.stability.Unstable;
+package com.xwiki.pickers.userList;
 
 /**
- * Picker representation for the color parameter of the status macro.
+ * Picker for the properties field of the user list macro.
  *
- * @version $Id$
  * @since 1.31.4
```
raphj:
```suggestion
 * @since 1.32.0
```

New APIs should trigger a new major versions.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Code/UserListConfig.xml
```diff
@@ -0,0 +1,981 @@
...
+#set ($inputName = "${className}_0_${propName}")
+#set ($inputId = "${configDoc.fullName}_${inputName}")
+#set ($currentValues = $obj.getProperty($propName).list)
+#set ($selectedValues = $obj.getProperty($propName).getValue())
+#set ($fieldLabel = $services.localization.render('userlist.config.bannedFields.label'))
+#set ($fieldDescription = $services.localization.render('userlist.config.bannedFields.label.description'))
+#set ($saveLabel = $services.localization.render('userlist.config.save.button'))
+
+{{html}}
+  &lt;form id="userlist_$escapetool.xml($configDoc.fullName)" method="post" action="$doc.getURL('save')" class="xform"&gt;
```
raphj, line 61: Maybe escape the action parameter value: `$escapetool.xml($doc.getURL('save'))`.

I see how this might be superfluous as it's supposed to be a fixed string.

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Code/UserListConfig.xml
```diff
@@ -0,0 +1,981 @@
...
+            &lt;option
+              value="$escapetool.xml($property.name)"
+              #if ($selectedValues.contains($property.name))
+                selected="selected"
+              #end&gt;
+              $escapetool.xml($services.localization.render("userlist.param.properties.$property.name"))
+            &lt;/option&gt;
+          #end
+        &lt;/select&gt;
+        &lt;input name="$inputName" type="hidden" value="$selectedValues"&gt;
```
raphj, line 69: Probably a bit paranoid, but `$selectedValues` should probably be escaped (even if it's supposed to only contain simple value, but I suppose a user could enter an arbitrary string in the configuration)

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Code/UserListConfig.xml
```diff
@@ -0,0 +1,981 @@
...
+        }).get();
+      }
+
+      function syncHidden() {
+        if (!hidden.length) {
+          return;
+        }
+        let values = getOptions()
+          .filter(function (o) { return o.selected; })
+          .map(function (o) { return o.value; });
```
raphj: This functional-style code could look a bit better with arrow functions :-)

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Code/UserListConfig.xml
```diff
@@ -0,0 +1,981 @@
...
+  background: transparent !important;
+  font-size: 14px;
+  color: #172b4d;
+}
+
+.mp-picker input.mp-search:focus {
+  border: none !important;
+  outline: none !important;
+  box-shadow: none !important;
+}
```
raphj: Make sure that it's still possible for a user navigating with the keyboard to notice what's focused.

(it's possibly the case, I haven't checked)

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Code/UserListConfig.xml
```diff
@@ -0,0 +1,981 @@
...
+        <size>1</size>
+        <unmodifiable>0</unmodifiable>
+        <values>currentPage|onDemand|always</values>
+        <classType>com.xpn.xwiki.objects.classes.StaticListClass</classType>
+      </use>
+    </class>
+    <property>
+      <cache>long</cache>
+    </property>
+    <property>
```
raphj: There are a lot of hard-coded colors in this CSS. Shouldn't we be using the theme-defined css variables (with of course a default for when they aren't defined?)

raphj (changes requested):

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
Farcasut: Mb. I was focused on adding the @ since to all the classes that I didn't realize that this class was not introduced by me

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
Farcasut: It will always be broken until https://jira.xwiki.org/browse/XWIKI-24578 is fixed. We could mention in the release that we are breaking the API

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Code/UserListConfig.xml
Farcasut: That is actually a really good catch. Someone could have written a custom string in the config and achieved a nice code injection. It would have required admin privileges, but still

xwiki-pro-macros-ui/src/main/resources/XWiki/Macros/Code/UserListConfig.xml
Farcasut: I've chosen to hardcode the values because I find the LESS variables lacking in this case. To achieve the UI that is in the video, I'd have to use weird combinations, like the border color for the background of the tag, to get the right amount of gray to have contrast, and I find that confusing both as a developer and as an end user. I think it would be a bad experience for an end user to modify the border color and have to update the tag color too, and I think they'd rather write a stylesheet. Maybe I could define my own LESS variables for this feature and link the defaults to the ones defined in the skin, but I'm not sure if it's worth the work. If you think it's worth it, I can start looking into how to write custom variables and display them in the skin customization section of the admin. Maybe I could define them as CSS variables to make overwriting easier. WDYT?

raphj: Definitely don't use LESS variables as we are moving out of them really soon. I would use CSS variables (with hard-coded defaults for XWiki 15 and 16).

I wouldn't define new CSS variables for customization. I would just use the CSS variables from XWiki standard, so your selector automatically matches the theme: https://www.xwiki.org/xwiki/bin/view/documentation/xs/dev/flamingo-skin/style-variables/use-css-properties/
What colors wouldn't be reachable from this? Note that you can darken and brighten colors (e.g. https://stackoverflow.com/questions/55329996/how-to-create-color-shades-using-css-variables-similar-to-darken-of-sass) if you really must (although I would refrain from doing so).

Last question: I haven't fully understood why you didn't use selectize?

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
raphj: I figured :rofl:

xwiki-pro-macros-api/src/main/java/com/xwiki/macros/userlist/macro/UserListMacroParameters.java
raphj: Ah, too bad...

xwiki-pro-macros-api/src/main/resources/templates/html_displayer/userlistproperties/edit.vm
```diff
@@ -35,22 +37,21 @@
...
+            {'name': $property.name, 'value': $property.name}
             ))
         #end
     #end
 #end
 #set ($usernamePropery = 'username')
 #if (!$bannedFields.contains($usernamePropery))
     #set ($discard = $parameters.get('options').add(
-        {'name': $usernamePropery, 'value': $usernamePropery}
+    {'name': $usernamePropery, 'value': $usernamePropery}
```
raphj: Since last time, these two lines have been weirdly deindented, I don't know if you meant to do this.

raphj (approved):
The big cleaning 😅

